PR #10, opened 2017-04-03
Resource limits jenkins

kbsingh:
@aslakknutsen this should help normalise the Jenkins resources; we setup a 90Mi claim as minimal to make sure small stuff like the content cache etc can work. Max'ing out Jenkins itself at 700Mi, it should not need anything more than that. Also moving the terminating=jobs to be 1Gi max, since most builds would not need this much ram.

@vpavlin once @aslakknutsen has this in, can you sync your account ( rebuild ? ) and validate.

### DIFF
--- a/template/fabric8-online-jenkins-openshift.yml
+++ b/template/fabric8-online-jenkins-openshift.yml
@@ -33,6 +33,37 @@ objects:
       secrets: "40"
       services: "5"
 - apiVersion: v1
+  kind: LimitRange
+  metadata:
+    name: resource-limits
+  spec:
+    limits:
+    - max:
+        cpu: "1368m"
+        memory: "700Mi"
+      min:
+        cpu: "17m"
+        memory: "90Mi"
+      type: Pod
+    - default:
+        cpu: "1"
+        memory: 512Mi
+      defaultRequest:
+        cpu: 60m
+        memory: 307Mi
+      max:
+        cpu: "1368m"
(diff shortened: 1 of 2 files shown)


template/fabric8-online-jenkins-openshift.yml
```diff
@@ -33,6 +33,37 @@ objects:
...
       services: "5"
 - apiVersion: v1
+  kind: LimitRange
+  metadata:
+    name: resource-limits
+  spec:
+    limits:
+    - max:
+        cpu: "1368m"
+        memory: "700Mi"
```
aslakknutsen: the max for a Terminating pod should be 1G?

rawlingsj:
@kbsingh @aslakknutsen we shouldn't be making changes here.  The templates are generated and released as part of fabric8-online.  Any changes should be a PR to https://github.com/fabric8io/fabric8-online.  One thing we need to do as part of the fabric8-online release is to PR on this repo with the new versions.  I'm planning on doing that later today.

kbsingh:
@aslakknutsen I've not tuned down the non-terminating resource ask yet, can we try this and see where we hit the bottlenecks ? I guess it will all boil down to what we need to burst upto and when.

aslakknutsen:
@rawlingsj Yes, I agree it should be done there.. But until we know what this actually should be it's a lot easier/quicker to make the changes here, then replicate upstream when done.

PR #11, opened 2017-04-03
New configuration to make Che work on new int environment

l0rd:
This setup has to be used when che workspaces are deployed as non-terminating pods (that's currently the only option).

And for the love of code duplication...https://github.com/fabric8io/fabric8-online/pull/118


kbsingh:
LGTM

PR #16, opened 2017-04-05
Update Che template to version 1.0.79

l0rd:


l0rd:
That has already been merged with commit 66459eaec99bcbebf8e27984003c815b88828a84

PR #24, opened 2017-04-20
Add resource requests/limits

vpavlin:
Please, let me know if you don't think these limits are reasonable.


aslakknutsen:
Hard to tell. We haven't really seen it under any form of load. :)

vpavlin:
Can we do any load testing? Or is that too artificial?

vpavlin:
We cannot got to prod without limits...

OpenShiftTemplate.yml
```diff
@@ -15,6 +15,21 @@ objects:
     selector:
       app: f8tenant
       deploymentconfig: f8tenant
+    strategy:
+      resources:
+        requests:
```
aslakknutsen: I thought OSO ppl said generally stay away from request. Set limits only?

vpavlin:
@aslakknutsen I've updated this - removed `requests` and lifted the limits a bit. Please review

PR #39, opened 2017-04-26
Update fabric8-online version to 1.0.122

fabric8cd:


rawlingsj:
For now we need to set the template parameter in the jenkins team namespace as well.  Let me do that quickly.

rawlingsj:
@aslakknutsen this PR should include the parameter in the jenkins online template for you to override, should be available in ~20 mins based on the previous release https://github.com/fabric8io/fabric8-online/pull/191

PR #54, opened 2017-04-27
Update fabric8-online version to 1.0.136

fabric8cd:


rawlingsj:
@aslakknutsen we need a better way to describe what fix / features are in each PR but we dont have that yet.  There is a fix to the bayesian plugin that is needed to work with the secure recommender endpoint, that is in fabric8/jenkins-openshift: v1a399cc  which is this 1.0.136 (and the previous few too).  We need this version to get the stack reports.

rawlingsj:
@aslakknutsen ah .137 has been raised so maybe you were waiting for that https://github.com/fabric8io/fabric8-init-tenant/pull/55

PR #88, opened 2017-05-23
make it easier to integrate into gofabric8 by not bringing in goa dependencies via the log.Debug statement - and also letting gofabric8 have more control over how log statements are formatted as templates are loaded from versions or files

jstrachan:


openshift/init_tenant.go
```diff
@@ -145,6 +144,7 @@ func do(config Config, callback Callback, username, usertoken string, templateVa
 // or default to the OOTB template included
 func loadTemplate(config Config, name string) ([]byte, error) {
 	teamVersion := config.TeamVersion
+	logCallback := config.LogCallback
```
aslakknutsen: Pseudo to avoid the nil check on each call..

```
config.GetLogger() {
   if c.LogCallback == nil {
    return NilLogger()
  }
}
```

aslakknutsen: And the log interface needs to be more 'advanced' to handle the way the InitiTenet function use it. (as you will need that callback function since it does most of the work)

jstrachan: thx, lemme try that

jstrachan: didn't grok your 2nd comment; how did you want the callback prototype to change?

aslakknutsen: https://github.com/fabric8io/fabric8-init-tenant/blob/master/controller/tenant.go#L234

jstrachan: so you mean you 'd rather the LogCallback take a map right? Or you mean to reuse the existing Callback (which was really to process the results of operations)

jstrachan: @aslakknutsen didn't quite grok the above ;)

aslakknutsen: You will need to reuse the InitTenant function for the tenant stuff to work at all. 

Meaning it you probably want to use the same logcallback function in there. But the function use a more extended logging then what you currently support in that function.

jstrachan: So I'm trying to avoid using the underlying almighty-core/log package as that brings in goa which then borks on jwt dependency clashes with kubernetes. It seems to work fine without that `InitTenant` controller callback & I could just replica whatever it does if I need to.

jstrachan: the only thing the controller `Callback` does is update the namespace in the database and do logging with almighty-core/log; which we don't need for the CLI tool (and the latter we can't build with ;)

aslakknutsen: Not at all.

That is the main logic of the whole tenant create/update. It reacts to the response codes from POST N and does the things it needs to to PUT or DELETE etc.. 

Without that you're reimplementing the init logic..

jstrachan: @aslakknutsen all I see here https://github.com/fabric8io/fabric8-init-tenant/blob/master/controller/tenant.go#L195-L243 is log statements and one call https://github.com/fabric8io/fabric8-init-tenant/blob/master/controller/tenant.go#L218 to service.UpdateNamespace() which I don't need. What am I missing? As I said without using any of this code the CLI is working fine for me (I don't want/need any DB updates)

aslakknutsen: 'init' will kinda work, but without that part it won't e.g. update resources that has the same name. Things that are auto setup by the ProjectRequest on os but need changes for us to use it. (some of these things are related to specific oso things, but still..) That function is the _main_ logic of the whole thing. Everything else is primarily plumming.

https://github.com/fabric8io/fabric8-init-tenant/blob/master/openshift/apply.go#L231

jstrachan: as I said, its working - all the namespaces and resources are created. Though `developer` can't create role bindings; but they get the namespaces, che, jenkins + content-repository

aslakknutsen: If so, which I doubt, is by luck, not intention.

jstrachan: @aslakknutsen how about we merge this as it is, we can then do testing of jenkins stuff in our CD pipeline before we release - then I'll look at refactoring to reuse the `controller.InitTenant` later on?

PR #90, opened 2017-05-26
allow host verification to be disabled via env var

jstrachan:
when running on minishift we don't yet have certs on routes so fabric8-init-tenant fails on startup to load from KC


jstrachan:
[test]

jstrachan:
[test]

jstrachan:
[test]

PR #93, opened 2017-05-26
Add optional env vars so we can provision users tenant apps on the sa…

rawlingsj:
…me cluster as the init tenant service

This means we can now run the fabric8 init service on minishift


jstrachan:
[test]

jstrachan:
[test]

jstrachan:
[test]

PR #145, opened 2017-07-10
Replace html/template with simple RegExp replace

aslakknutsen:
Avoids templates being html/xml safe escaped

### DIFF
--- a/openshift/process.go
+++ b/openshift/process.go
@@ -1,24 +1,21 @@
 package openshift
 
-import (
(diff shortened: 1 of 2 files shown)


fabric8cd:
@aslakknutsen snapshot fabric8-init-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-init-tenant:SNAPSHOT-PR-145-1`

PR #152, opened 2017-07-13
lets allow the exposer to be configured via env var on k8s

jstrachan:
support vanilla kubernetes for init tenant


fabric8cd:
@jstrachan snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-152-1`

PR #158, opened 2017-07-13
lets pass in the keycloak URL for kubernetes jenkins

jstrachan:


fabric8cd:
@jstrachan snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-158-1`

PR #170, opened 2017-07-18
Split init and update strategy

### LINKED ISSUE #167: Use different strategies for Init vs Update
We moved to use 'oc apply' due to update issues with the pure brute force POST/DELETE method, but that  takes the init tenant time from about 400ms to 2:50 min. This causes some timing issues down the lines in e.g. the Forge Wizard that require the Jenkins Services to be around to discover the Jenkins Master URL to post as a GitHub WebHook Callback.

Split the Init and Update strategies, using pure POST/DELETE on Init and oc apply on Update.

aslakknutsen:
Rely on Async REST method to init Che and Jenkins,
while using oc apply on Che and Jenkins update.

REST method is a lot faster, and since there is nothing
to update at this point we can do a simple brute force POST/DELETE/POST
session.

* Removed Reorder of templates since they now come ordered from upstream.

Fixes #167

### DIFF
--- a/controller/tenant.go
+++ b/controller/tenant.go
@@ -132,7 +132,7 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
 		if openshift.KubernetesMode() {
 			oc = userConfig
 		}
-		err = openshift.InitTenant(
+		err = openshift.UpdateTenant(
 			ctx,
 			c.keycloakConfig,
 			oc,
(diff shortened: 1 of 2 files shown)


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-170-1`

PR #235, opened 2017-08-07
Add Keycloak Broker URLs in template variables

### LINKED ISSUE #228: Init tenant should set the correct Keycloak URL
In fabric8-online[1] we have introduced two new configs for che-server:

- keycloak-oso-endpoint
- keycloak-github-endpoint

These are currently hardcoded in fabric8io configuration with KC production URLs. That's an issue because we may need to use a distinct instance of KC based on which cluster Che will be deployed to (prod / prod-preview). These 2 config maps should be configured by the tenant init script as it's already done for hostname-http

[1] https://github.com/fabric8io/fabric8-online/blob/master/apps/che/src/main/fabric8/cm.yml#L20-L21

aslakknutsen:
Fixes #228


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-235-1`

fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-235-2`

PR #253, opened 2017-08-11
Lock down access to the users namespaces

aslakknutsen:
* DELETE admin role in namespce 'user' on init

All other changes are reflected in the templates.

Update now assumes namespaces will be there and will not
recreate them. Each namespace is updated via oc apply in parallel.


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-253-1`

PR #283, opened 2017-08-18
Replace all non a-z0-9 chars in username to use as basename

aslakknutsen:
Related to openshiftio/openshift.io#619


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-283-1`

PR #287, opened 2017-08-23
Add +1 addition parameter to teams template test

aslakknutsen:
Fabric8-Maven-Plugin does not separate between the parameters
required in the Openshift and the K8 templates. Even tho
the new parameter is not used in the Openshfit templates it's
defined in the template spec.


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-287-1`

PR #443, opened 2017-11-09
Use same KC URL for Tenant and Templates

aslakknutsen:
Related to openshiftio/openshift.io#1303


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-443-1`

Issue #444, opened 2017-11-09
Reset environment / update tenant is broken for che


ibuziuk:
looks like `defaultAuthURL` is wrong in the tenant configuration -
https://github.com/fabric8-services/fabric8-tenant/blob/master/configuration/configuration.go#L355

navigating to che server route results in 404

ibuziuk:
Sorry, this is not the case - I have forgotten to clear values on tenant page before resetting environment

PR #445, opened 2017-11-09
fabric8-tenant #444: Setting correct defaultAuthURL

ibuziuk:


fabric8cd:
@ibuziuk snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-445-1`

configuration/configuration.go
```diff
@@ -352,7 +352,7 @@ const (
 	devModeKeycloakURL   = "https://sso.prod-preview.openshift.io"
 	devModeKeycloakRealm = "fabric8-test"
 
-	defaultAuthURL                  = "https://auth.prod-preview.openshift.io"
+	defaultAuthURL                  = "https://sso.prod-preview.openshift.io/auth"
```
alexeykazakov: Why do you need Keycloak URL as Auth URL? It won't work.
See https://github.com/fabric8-services/fabric8-tenant/blob/master/main.go#L134
If you use KC URL here you will get:
`KEYCLOAK_GITHUB_ENDPOINT = https://sso.prod-preview.openshift.io/auth/api/token?for=https://github.com`
by default. Which won't work.
It should be
`KEYCLOAK_GITHUB_ENDPOINT = https://auth.prod-preview.openshift.io/api/token?for=https://github.com` instead.

ibuziuk: @alexeykazakov not sure how this should be handled on init-tenant side, but the correct value for che defined in this PR - https://github.com/fabric8-services/fabric8-tenant-che/pull/60/files

Also, with current config init teant / reset environment is completely broken for che - https://github.com/fabric8-services/fabric8-tenant/issues/444

ibuziuk: actually #444 does not really exist (I have forgotten to clear tenant values before resetting env.) but che on prod-preview is still broken (403 when navigating to the route)

alexeykazakov: Then I guess you actually need to set `defaultKeycloakURL = "https://sso.prod-preview.openshift.io/auth"` instead of `"https://sso.prod-preview.openshift.io"` in https://github.com/fabric8-services/fabric8-tenant/blob/f5a9ffa4869d9857ef757c204f7b2465cb8afef8/configuration/configuration.go#L346

aslakknutsen:
Default Auth URL should point to auth service. Che is getting KC URL https://github.com/fabric8-services/fabric8-tenant/blob/master/configuration/configuration.go#L337

PR #450, opened 2017-11-09
pcp/prometheus implant: enable collection of intra-pod data for monitoring

fche:
This patch adds two pcp-related background processes to the container,
costing about 25MB more memory.  They allow collection of intra-pod
system stats, plus prometheus metrics at http://localhost:8080/metrics
for relay to a central logging server.  They are passive (use zero
CPU) unless such a connection is made.

(A predecessor of the same patch has long existed in fabric8-wit, and the
same one is pending in a PR there.)


aslakknutsen:
[test]

aslakknutsen:
[test]

PR #466, opened 2017-11-17
Expose CHE_MULTITENANT_SERVER and OSIO_TOKEN to templates

aslakknutsen:
Related to redhat-developer/rh-che#377

PR #473, opened 2017-11-21
Add `requestId` and `identityId` to the template variables for the Che migration job POD

davidfestal:
These 2 values should be added to the config map that will be read by the run-once POD that will do the single-tenant to multi-tenant migration.

This PR is a followup of this one: https://github.com/fabric8-services/fabric8-tenant-che/pull/74/files and should be merged one the latter is merged

### DIFF
--- a/openshift/process_template.go
+++ b/openshift/process_template.go
@@ -11,7 +11,7 @@ import (
 	"github.com/fabric8-services/fabric8-tenant/toggles"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
 	jwt "github.com/dgrijalva/jwt-go"
-    "github.com/fabric8-services/fabric8-wit/log"
+	"github.com/fabric8-services/fabric8-wit/log"
 )
 
 type FilterFunc func(map[interface{}]interface{}) bool
--- a/openshift/process_template.go
+++ b/openshift/process_template.go
@@ -10,7 +10,8 @@ import (
 
 	"github.com/fabric8-services/fabric8-tenant/toggles"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
-	"github.com/fabric8-services/fabric8-wit/log"
+	jwt "github.com/dgrijalva/jwt-go"
+    "github.com/fabric8-services/fabric8-wit/log"
 )
 
 type FilterFunc func(map[interface{}]interface{}) bool
@@ -136,7 +137,7 @@ func LoadProcessedTemplates(ctx context.Context, config Config, username string,
 		if token != nil {
 			vars["OSIO_TOKEN"] = token.Raw
 			id := token.Claims.(jwt.MapClaims)["sub"]
-			if token != nil {
+			if id != nil {
 				vars["IDENTITY_ID"] = id
 			}
 		}
--- a/openshift/process_template.go
+++ b/openshift/process_template.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/fabric8-services/fabric8-tenant/toggles"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
+	"github.com/fabric8-services/fabric8-wit/log"
 )
 
 type FilterFunc func(map[interface{}]interface{}) bool
@@ -134,7 +135,12 @@ func LoadProcessedTemplates(ctx context.Context, config Config, username string,
 		token := goajwt.ContextJWT(ctx)
 		if token != nil {
 			vars["OSIO_TOKEN"] = token.Raw
+			id := token.Claims.(jwt.MapClaims)["sub"]
+			if token != nil {
+				vars["IDENTITY_ID"] = id
+			}
 		}
+		vars["REQUEST_ID"] = log.ExtractRequestID(ctx)
 		cheType = "mt-"
 	}
 
--- a/openshift/process_template.go
+++ b/openshift/process_template.go
@@ -12,6 +12,7 @@ import (
 	"github.com/fabric8-services/fabric8-tenant/toggles"
 	"github.com/fabric8-services/fabric8-wit/log"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
+	"github.com/pkg/errors"
 )
 
 type FilterFunc func(map[interface{}]interface{}) bool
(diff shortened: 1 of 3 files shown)


openshift/process_template.go
```diff
@@ -134,7 +135,12 @@ func LoadProcessedTemplates(ctx context.Context, config Config, username string,
 		token := goajwt.ContextJWT(ctx)
 		if token != nil {
 			vars["OSIO_TOKEN"] = token.Raw
+			id := token.Claims.(jwt.MapClaims)["sub"]
```
ibuziuk: https://github.com/fabric8-services/fabric8-wit/blob/master/log/context.go#L14 ?

davidfestal: well, I didn't use the `extractIdentityID` function since we already have the token, and we don't need an error (which is why a tuple is returned by `extractIdentityID`)

openshift/process_template.go
```diff
@@ -134,7 +135,12 @@ func LoadProcessedTemplates(ctx context.Context, config Config, username string,
...
+			if token != nil {
```
l0rd: `if id != nil`?

davidfestal: sure :-/

openshift/process_template.go
ibuziuk: ok, makes sense

xcoulon: so what happens when you don't have the token ? I don't see why you wouldn't re-use this utility function.

openshift/process_template.go
```diff
@@ -10,6 +10,8 @@ import (
 
 	"github.com/fabric8-services/fabric8-tenant/toggles"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
+	jwt "github.com/dgrijalva/jwt-go"
+    "github.com/fabric8-services/fabric8-wit/log"
```
xcoulon: formatting looks incorrect here

openshift/process_template.go
davidfestal: I just don't set it.
I can reuse the utility, but it will do the JWT token extraction twice then.

xcoulon: ok

davidfestal:
@xcoulon it seems no PR build was run after my last commit. any idea ?

alexeykazakov:
[test]

openshift/process_template.go
```diff
@@ -134,7 +136,12 @@ func LoadProcessedTemplates(ctx context.Context, config Config, username string,
 		token := goajwt.ContextJWT(ctx)
 		if token != nil {
 			vars["OSIO_TOKEN"] = token.Raw
+			id := token.Claims.(jwt.MapClaims)["sub"]
+			if id != nil {
```
alexeykazakov: If there is no "sub"/ID in the token then there is something fundamentally broken... You should return an error rather than silently ignore it.

fabric8cd:
@davidfestal snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-473-3`

openshift/process_template.go
ibuziuk: in this IMO it would be better to reuse existing API (we can leave with double token extraction)

fabric8cd:
@davidfestal snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-473-4`

openshift/process_template.go
davidfestal: @ibuziuk 

I can do it if you like, but then I'll have to check for the text of the error, to throw an error if `sub` is not there, but *not throw an error* when the token itself is not there ? We shouldn't throw an error if the token itself is not there.

Seems to me that this would make the code even more complicated, no ?

FTR the code of the function I would reuse is:

```
func extractIdentityID(ctx context.Context) (string, error) {
	token := goajwt.ContextJWT(ctx)
	if token == nil {
		return "", errors.New("Missing token")
	}
	id := token.Claims.(jwt.MapClaims)["sub"]
	if id == nil {
		return "", errors.New("Missing sub")
	}

	return id.(string), nil
}
```

davidfestal: @alexeykazakov 

> If there is no "sub"/ID in the token then there is something fundamentally broken... You should return an error rather than silently ignore it.

I'm not sure it's the place to make such a check and throw such an error, but if you think it is I can throw an error when `id == nil`. Should I throw an exception ? FYI I'm a Go newbie, so if you could provide me with the piece of code you would add to correctly manage this error, that would be great :-)

ibuziuk: `to throw an error if sub is not there, but not throw an error when the token itself is not there ? We shouldn't throw an error if the token itself is not there.`

is it possible scenario when context is missing token, or it also means that something is broken ?

davidfestal: I assume that if the tenant update if run by some sort of batch command (for a whole group of users for example), then the Keycloak token coming from the user http request will not be there.

ibuziuk: ok, whatever works than ;-)

davidfestal: > If there is no "sub"/ID in the token then there is something fundamentally broken... You should return an error rather than silently ignore it.

I now return an error in this case. Thanks @xcoulon for your help.

fabric8cd:
@davidfestal snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-473-6`

davidfestal:
@xcoulon could you merge this PR ? The last time such a PR was done (https://github.com/fabric8-services/fabric8-tenant/pull/466), @aslakknutsen had to then create a new PR to fix tests (increase the template parameter numbers): https://github.com/fabric8-services/fabric8-tenant/pull/467.

I assume we will have to do the same this time, since there are 2 more template parameters.

davidfestal:
so @xcoulon @alexeykazakov do you think we could merge this one ?

fabric8cd:
@davidfestal snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-473-8`

alexeykazakov:
[test]

Issue #229, opened 2017-08-07
Support for Early access / non updateble accounts


aslakknutsen:
Certain OSiO accounts need early access to tenant deployments for 'pre live' verification, e.g. as part of EE tests, while other dev accounts would only want manual updates e.g. devs testing in prod.

pbergene:
Are these all accounts that need access or do they already have? 
I only changed permission of the relevant namespaces - I can guess -che for some of the above, but do we have this information?  In that case, what do you think about selectively updating parts of the tenant setup - I guess there might be cases where you'd want the latest version of the parts you are not working on. 

To complement the list: sgehwolf-jenkins: sgehwolf

chmouel:
For the build team we would need : 

@sthaha 
@rupalibehera 
@chmouel
@jaseemabid 
@hrishin 

Thanks,

PR #495, opened 2017-12-08
Use whitelist to restrict permission to user profile override

xcoulon:
Using the tenant service with a feature that MUST be named
`user.update-tenant` and which should be configured with
the user ids, in order to restrict permission to use the content
of `contextInformation.tenantConfig` in the user's profile in `auth`
to change the default versions of Che/Jenkins/etc.

Include a `Toggle Client` that can use a mock unleashclient for testing,
as well as an http client in the controller for mocking the interactions
with `auth`.


fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-495-1`

controller/tenant.go
```diff
@@ -175,22 +185,30 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
+	if len(c.authURL) > 0 && c.toggleClient.IsEnabled(token, toggles.UserUpdateTenantFeature, false) {
+		log.Info(ctx, map[string]interface{}{"auth_url": c.authURL, "http_client_transport": reflect.TypeOf(c.httpClient.Transport)}, "retrieving user's profile...")
+		authClient, err := newAuthClient(ctx, c.httpClient, c.authURL)
 		if err != nil {
+			log.Error(ctx, map[string]interface{}{"auth_url": c.authURL}, "unable to parse auth URL")
 			return config, err
 		}
-		if status < 200 || status > 300 {
-			return config, fmt.Errorf("Failed to GET url %s due to status code %d", url, status)
+		resp, err := authClient.ShowUser(ctx, auth.ShowUserPath(), nil, nil)
```
alexeykazakov: Close response body in defer and make sure the body has been read. Otherwise you will leak FDs.

xcoulon: oh yes, indeed! thanks, @alexeykazakov!

xcoulon:
[test]

controller/tenant.go
```diff
@@ -198,13 +198,16 @@ func (c *TenantController) loadUserTenantConfiguration(ctx context.Context, conf
...
 		}
 		resp, err := authClient.ShowUser(ctx, auth.ShowUserPath(), nil, nil)
+		defer resp.Body.Close()
+
 		if err != nil {
 			log.Error(ctx, map[string]interface{}{"auth_url": auth.ShowUserPath()}, "unable to get user info")
 			return config, errs.Wrapf(err, "failed to GET url %s due to error", auth.ShowUserPath())
 		}
 		if resp.StatusCode < 200 || resp.StatusCode > 300 {
 			return config, fmt.Errorf("failed to GET url %s due to status code %d", resp.Request.URL, resp.StatusCode)
```
alexeykazakov: You need to read the body here too because otherwise it may leak FD even if you close the body :)
General rule for making sure we don't have a FD leak - If err!=nil then read the body (once) and close it.

controller/tenant.go
```diff
@@ -175,22 +185,33 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
-	authHeader := token.Raw
-	if len(c.usersURL) > 0 {
-		url := strings.TrimSuffix(c.usersURL, "/") + "/user"
-		status, body, err := openshift.GetJSON(config, url, authHeader)
+// loadUserTenantConfiguration loads the tenant configuration for `auth`,
+// allowing for config overrides based on the content of his profile (in auth) if the user is allowed
+func (c *TenantController) loadUserTenantConfiguration(ctx context.Context, config openshift.Config) (openshift.Config, error) {
+	token := goajwt.ContextJWT(ctx)
+	// use the toggle service to verify that the user is allowed to change is profile based on the settings passed in the URL
+	if len(c.authURL) > 0 && c.toggleClient.IsEnabled(token, toggles.UserUpdateTenantFeature, false) {
```
aslakknutsen: This is not really a good use of Toggles. Toggles != Permanent User config.

https://github.com/fabric8-services/fabric8-tenant/issues/229

controller/tenant.go
xcoulon: ok, let me move this line just after `js, err := simplejson.NewFromReader(resp.Body)`, then. I assume this will respond to the golden rule you explained, right ?

controller/tenant.go
xcoulon: Sorry, I'm not following you here: the goal of the change is to allow (or not) user config override, based on the toggle state for the user. The user's tenants will be initialised with the default versions for Che/Jenkins/etc, unless toggle is enabled and user has some specific settings in his/her profile.

fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-495-3`

controller/tenant.go
aslakknutsen: @xcoulon Yes, I get that. But this is not a temporary setting, right? It's not related to 'let's try something out' or 'while we deploy this new feature' type of setting. It's a permanent feature of the user that he has more rights or whatever. Since it's permanent it's the wrong use of the Toggle Service.

xcoulon: @aslakknutsen ok, got it, so using the toggle service is not the right way to handle this permission. So, should we set a flag in the user's profile or use `auth` to check the permission to override the profile ?

aslakknutsen: @xcoulon I would add some type of 'group' to the "tenant" table. That way we can use the info when doing automatic updates as well. E.g. roll out to group N first, then wait 24h and unless stopped, roll out to group X.. etc etc

xcoulon: @aslakknutsen not following your idea: we need to restrict permission to override the default config to a list of users who would need to test PRs. How do you expect to handle this in the `tenant` table and how do you expect to manage the data (I'm thinking about an admin UI, here)

xcoulon: @aslakknutsen what about using an attribute in KC records to handle a whitelist of users who are allowed to override the config ?

aslakknutsen: @xcoulon The UserProfile service or what ever would mark a user to belong to some group (some service account is allowed to change that data on behalf of a user). 

The problem with using KC for this is that it would require a user token to look it up... which causes issues for auto update and extra pressure on KC.

xcoulon: what auto-update ? I thought this would be managed by an admin ?
if KC is not a good candidate, what about rules handled by our `auth` service, then ?

controller/tenant.go
alexeykazakov: If response is not a JSON (if it's not 200) then simplejson.NewFromReader(resp.Body) will fail I guess. So, instead do the following:
```
defer resp.Body.Close()
if err != nil {
   // return error
}
body := ioutil.ReadAll(resp.Body)
if resp.StatusCode < 200 || resp.StatusCode > 300 {
   // return error
}

// Unmarshal JSON from body.
```

Or

```
defer resp.Body.Close()
if err != nil {
   // return error
}
if resp.StatusCode < 200 || resp.StatusCode > 300 {
   ioutil.ReadAll(resp.Body)
   // return error
}

js, err := simplejson.NewFromReader(resp.Body)
```

Or simply use `defer rest.CloseResponse(resp.Body)` from https://github.com/fabric8-services/fabric8-auth/blob/master/rest/url.go#L46 ;)

controller/tenant.go
aslakknutsen: @xcoulon Auto update of tenants.

controller/tenant.go
xcoulon: ok, thanks for the explanations, @alexeykazakov. Am I wrong or does 
````
func CloseResponse(response *http.Response) {
	ioutil.ReadAll(response.Body)
	response.Body.Close()
}
````
reads the body but does not return it ? I thought we should read it once only ? Or at least once ? (In Java you cannot read the streams multiple times)

File: controller/tenant.go (hunk 3 above)
Comment (xcoulon) [reply controller/tenant.go]: @alexeykazakov @aslakknutsen ok, so what solution should I use in the mean time (i.e., until auth supports groups/team permissions), then ?

File: controller/tenant.go (hunk 2 above)
Comment (xcoulon) [reply controller/tenant.go]: fixed in https://github.com/fabric8-services/fabric8-tenant/pull/495/commits/06ade20f0fd51f6c2508612e6cff20f915b9eaa5

Comment (xcoulon) [reply controller/tenant.go]: hopefully ;)

**Comment by fabric8cd** [comment]
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-495-4`

File: controller/tenant.go (hunk 2 above)
Comment (alexeykazakov) [reply controller/tenant.go]: That's right. It reads the stream to make sure it's empty and then closes the body. If stream is already empty (body has been read) that's OK. It doesn't do any harm.

File: controller/tenant.go (hunk 3 above)
Comment (alexeykazakov) [reply controller/tenant.go]: We could enable custom jenkins/che only for specific users (by UUID) for now.

File: controller/tenant.go [hunk 4]
```diff
@@ -175,22 +186,36 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
+		log.Info(ctx, map[string]interface{}{"auth_url": c.authURL, "http_client_transport": reflect.TypeOf(c.httpClient.Transport)}, "retrieving user's profile...")
+		authClient, err := newAuthClient(ctx, c.httpClient, c.authURL)
 		if err != nil {
+			log.Error(ctx, map[string]interface{}{"auth_url": c.authURL}, "unable to parse auth URL")
 			return config, err
 		}
-		if status < 200 || status > 300 {
-			return config, fmt.Errorf("Failed to GET url %s due to status code %d", url, status)
+		resp, err := authClient.ShowUser(ctx, auth.ShowUserPath(), nil, nil)
+		if err != nil {
```
alexeykazakov: You forgot to remove lines 202-205 I guess (duplicates 208-211)

controller/tenant.go
```diff
@@ -175,22 +186,36 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
+			log.Error(ctx, map[string]interface{}{"auth_url": auth.ShowUserPath()}, "unable to get user info")
+			return config, errs.Wrapf(err, "failed to GET url %s due to error", auth.ShowUserPath())
+		}
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
```
alexeykazakov: Assuming you will remove lines 202-205 you may get a panic here because resp can be nil here.

controller/tenant.go
xcoulon: no, these 2 errors are different: `err` on line `201` comes from `authClient.ShowUser()` (i.e, something wrong happened while calling the auth service), whereas error on line `207` comes from reading the response body.

controller/tenant.go
xcoulon: See my comment above: I should keep both blocks.

controller/tenant.go
xcoulon: yes, that's the point of this issue and using the toggle service was one way to implement it (i.e, verify that the user's ID belongs to a whitelist). However, I understand @aslakknutsen's argument that toggle service should be used to enable/disable features that are not released yet, which is not the use-case for this solution. Hence my question: where do we keep track of the permission ?

controller/tenant.go
alexeykazakov: Oh, you are right! Looks good to me now.

controller/tenant.go
```diff
@@ -175,30 +182,70 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
-			return config, fmt.Errorf("Failed to GET url %s due to status code %d", url, status)
-		}
-		js, err := simplejson.NewJson([]byte(body))
-		if err != nil {
-			return config, err
+// loadUserTenantConfiguration loads the tenant configuration for `auth`,
+// allowing for config overrides based on the content of his profile (in auth) if the user is allowed
+func (c *TenantController) loadUserTenantConfiguration(ctx context.Context, config openshift.Config) (openshift.Config, error) {
+	// restrict access to users with a `featureLevel` set to `internal`
+	log.Info(ctx, map[string]interface{}{"auth_url": c.authURL, "http_client_transport": reflect.TypeOf(c.httpClient.Transport)}, "retrieving user's profile...")
```
aslakknutsen: Could you move the Client call/setup/parsing etc into some function? user, err := GetCurrentUser(ctx) or something.

xcoulon: good point, let me do that

fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-495-13`

controller/tenant.go
```diff
@@ -175,34 +182,84 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{"auth_url": c.authURL}, "unable to parse auth URL")
+		return nil, err
+	}
+	resp, err := authClient.ShowUser(ctx, auth.ShowUserPath(), nil, nil)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{"auth_url": auth.ShowUserPath()}, "unable to get user info")
+		return nil, errs.Wrapf(err, "failed to GET %s due to error", auth.ShowUserPath())
+	}
+	if err != nil {
```
aslakknutsen: ... just in case an err is set since the previous check?

xcoulon: yes ;) joke aside, let me remove that check!

fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-495-14`

PR #506, opened 2017-12-19
Add type Job to REST mapping

aslakknutsen:
Missing type caused init to fail on Unknown Type.
Add Type check in test suite.


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-506-1`

PR #510, opened 2017-12-28
Add support to talk to auth

surajssd:
- [x] Design the cluster token service
- [x] Design the service token service
- [x] Use the auth client to implement the above.
- [x] In main.go , get the service token from Auth and initialize the TenantController object
- [x] In Setup() controller , call the User Api to get the user's cluster info, and then get the corresponding cluster token.
- [x] Use the cluster token to talk to the openshift online API.
- [x] Decrypt stored cluster tokens


configuration/configuration.go
```diff
@@ -373,4 +391,8 @@ const (
 
 	defaultWitURL     = "https://api.prod-preview.openshift.io/api/"
 	defaultTogglesURL = "http://f8toggles/api"
+
+	varAuthGrantType = "AUTH_GRANT_TYPE"
```
sbose78: Since these are not config, but just constant name, do a `const` declaration of these in the relevant package.

osioauth/tokens.go
```diff
@@ -0,0 +1,129 @@
+package osioauth
```
sbose78: you can call the package `token` instead of `osioauth/tokens`
( in general avoid `osio` in naming because its a jargon of our downstream name :) )

osioauth/tokens.go
```diff
@@ -0,0 +1,129 @@
...
+	Detail string `json:"detail,omitempty"`
+	Status string `json:"status,omitempty"`
+	Title  string `json:"title,omitempty"`
+}
+
+type errorResponse struct {
+	Errors []authEerror `json:"errors,omitempty"`
+}
+
+func GetAuthAccessToken() (string, error) {
```
sbose78: Code to an interface.

have something like a 

```
type TokenService interface{
     GetAuthAccessToken() (string, error) 
     GetOpenShiftToken() (string,error)
}

type TokenClient struct{}

func (c *TokenClient)GetAuthAccessToken() (string, error) {
   // implement the method.
}
```

This will help you inject the service wherever needed, and also mock it in tests.

osioauth/users.go
```diff
@@ -0,0 +1,74 @@
...
+
+type userData struct {
+	Data struct {
+		Attributes struct {
+			Cluster string `json:"cluster,omitempty"`
+		}
+	}
+}
+
+func GetUserCluster(userID string) (string, error) {
```
sbose78: Same as above, add a `UserService` ..

osioauth/tokens.go
sbose78: example: https://github.com/fabric8-services/fabric8-auth/blob/master/token/keycloak/external_token.go#L26

sbose78:
You would need to add in the code to get the cluster token from Auth.

You could either get the OSIO service account token from auth when the app comes up ( inside main.go ) and initialize the `ClusterTokenClient` with the osio service account token.

```
type ClusterTokenService interface{
         Get(cluster string) ( string,error )
}

type ClusterTokenClient struct{
        AuthServiceAccountToken string
        ....
}

func (c *ClusterTokenClient) Get(cluster string) {
        // use c.AuthServiceAccountToken
}
```

Or do it everytime you make a call to get the cluster token.

osioauth/tokens.go
surajssd: done!

osioauth/tokens.go
surajssd: this I am still figuring out

configuration/configuration.go
surajssd: done

token/tokens_test.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+		{
+			name:    "make code fail on parsing output",
+			wantErr: true,
+			output:  "foobar",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
surajssd: @sbose78 let me know how does this approach of testing looks to you?

surajssd:
@sbose78  I have few questions wrt  mocking the tests (I have never done mocking before with golang)

so to mock the code idiomatically we have to make the input to those methods parameterized right?  Now I am not sure what all things should ideally be parameterized? I have done lot of research today reading about mocking, where I found multiple approaches, one being using httptest package which I have used to test the method (this has no mocking)

But for doing mock how do I go forward? Any pointers will help

token/tokens.go
```diff
@@ -0,0 +1,156 @@
...
+
+type ClusterTokenClient struct {
+	AuthServiceAccountToken string
+}
+
+func (c *ClusterTokenClient) Get(config *configuration.Data) error {
+	payload := strings.NewReader("grant_type=" + config.GetAuthGrantType() + "&client_id=" +
+		config.GetAuthClientID() + "&client_secret=" + config.GetClientSecret())
+
+	req, err := http.NewRequest("POST", config.GetAuthURL()+"/api/token", payload)
```
alexeykazakov: Instead of using a regular http client it's better to use Auth client. See an example here - https://github.com/fabric8-services/fabric8-wit/blob/master/auth/resource.go#L55
You will need to generate the auth client package in make for that - https://github.com/fabric8-services/fabric8-wit/blob/master/Makefile#L248

configuration/configuration.go
```diff
@@ -51,6 +51,10 @@ const (
 	varAPIServerInsecureSkipTLSVerify  = "api.server.insecure.skip.tls.verify"
 	varLogLevel                        = "log.level"
 	varLogJSON                         = "log.json"
+
+	varAuthGrantType = "AUTH_GRANT_TYPE"
```
sbose78: remove this.

configuration/configuration.go
```diff
@@ -51,6 +51,10 @@ const (
...
+	varAuthClientID  = "AUTH_CLIENT_ID"
```
sbose78: so something like https://github.com/fabric8-services/fabric8-notification/pull/53/files#diff-786c8376feaa68e9ebee88ca950d2774R25

token/tokens.go
```diff
@@ -0,0 +1,156 @@
...
+	if z.OpenShiftToken, err = parseToken(body); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func parseToken(data []byte) (string, error) {
+	// this struct is defined to obtain the accesstoken from the output
+	type authAccessToken struct {
```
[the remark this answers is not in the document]
surajssd: this is not used anywhere else so scoped it in that function only.

token/tokens.go
surajssd: as per shoubhik's suggestion, I will park this for a little while and make those changes as soon as it is usable or working!

surajssd: but yes noted

configuration/configuration.go
surajssd: done

configuration/configuration.go
surajssd: done

token/tokens.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (z *OpenShiftTokenClient) Get(config *configuration.Data, accessToken string, cluster string) error {
+	// auth can return empty token so validate against that
+	if accessToken == "" {
+		return fmt.Errorf("access token can't be empty")
+	}
+
+	// check if the cluster is empty
+	if cluster == "" {
+		return fmt.Errorf("cluster URL can't be empty")
```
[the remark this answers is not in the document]
surajssd: done

token/tokens.go
```diff
@@ -0,0 +1,156 @@
...
+
+type OpenShiftTokenService interface {
+	Get(config *configuration.Data, accessToken string, cluster string) error
+}
+
+type OpenShiftTokenClient struct {
+	OpenShiftToken string
+}
+
+func (z *OpenShiftTokenClient) Get(config *configuration.Data, accessToken string, cluster string) error {
```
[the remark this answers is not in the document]
surajssd: done

token/tokens.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return nil
+}
+
+type OpenShiftTokenService interface {
+	Get(config *configuration.Data, accessToken string, cluster string) error
+}
+
+type OpenShiftTokenClient struct {
+	OpenShiftToken string
```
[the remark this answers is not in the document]
surajssd: done

configuration/configuration.go
```diff
@@ -265,6 +285,10 @@ func (c *Data) GetAuthURL() string {
 	return c.v.GetString(varAuthURL)
 }
 
+func (c *Data) SetAuthURL(value string) {
```
sbose78: not needed, if you need to set in tests, then you can do an `os.SetEnv(..)` and then re-initialize the config object

token/tokens.go
sbose78: Your code is in better shape now, please go ahead and start using the Auth client.

sbose78:
TODO:

1. In `main.go` , get the service token from Auth and initialize the `TenantController` object
2. In your `Setup()` controller , call the User Api to get the user's cluster info, and then get the corresponding cluster token.
3. Use the cluster token to talk to the openshift online API.

main.go
```diff
@@ -132,7 +113,7 @@ func main() {
 	}
 
 	templateVars["KEYCLOAK_URL"] = ""
```
sbose78: Why do we need `templateVars` ? 
In general, it isn't a good practice to use free form maps. You might need to wrap it in a `struct`

token/users.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+type UserProfileService interface {
+	GetUserCluster(userID string) (string, error)
+}
+
+type UserController struct {
+	Config                     *configuration.Data
+	serviceAccountTokenService ServiceAccountTokenService
```
sbose78: The Users API is open and doesn't really need a service account token :)

token/users.go
```diff
@@ -0,0 +1,66 @@
...
+			Cluster string `json:"cluster,omitempty"`
+		}
+	}
+}
+
+type UserProfileService interface {
+	GetUserCluster(userID string) (string, error)
+}
+
+type UserController struct {
```
sbose78: Call it a `UserProfileClient` . We reserve the usage of `Controller`.. only for the Goa request controllers.

sbose78:
We avoid using  package/file names having `utils` because it starts looking like we didn't know where to put it. Could you move the methods to the move appropriate location? Feel to free to create meaningful packages/sub-packages.

sbose78:
As discussed, please start consuming the auth goa client.

token/tokens.go
surajssd: this is done!

configuration/configuration.go
surajssd: removed, done!

token/users.go
surajssd: removed!

token/users.go
surajssd: done!

main.go
surajssd: not sure how to handle this!

aslakknutsen: @sbose78 They are 'free' variables used to fill in the Openshift templates. `oc process -p`

PR #516, opened 2018-01-27
Call /api/users/id to avoid endless loop with auth

aslakknutsen:
auth /api/user calls tenant setup calls auth /api/user...

Related to openshiftio/openshift.io#1687


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-516-1`

PR #517, opened 2018-01-27
Expose basic go runtime metrics via Prometheus

aslakknutsen:
Metrics are exposed under /metrics and the used port
can be controlled via F8_METRICS_HTTP_ADDRESS. By
default same port as F8_HTTP_ADDRESS is used.


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-517-1`

PR #522, opened 2018-01-31
Use whitelist to restrict permission to user profile override

aslakknutsen:
Only certain users should have access to override the tenant
profile in production. Locked down to the feature group internal.

Fixes #495


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-522-1`

xcoulon:
cool @aslakknutsen, I see you used parts of #495 to fix it once for all :)

PR #524, opened 2018-01-31
Add better equality check for Cluster URL between User and Cluster API

aslakknutsen:
Related to openshiftio/openshift.io#2122


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-524-1`

PR #541, opened 2018-02-27
Disable 'pullForce=true' when retrieving cluster SA token (OSIO#2304)

xcoulon:
`auth-service` stores the clusters' service account tokens in an encrypted
form and it cannot decrypt them (only `tenant-service` can do it), so it
is not able to validate the tokens. This will have to be done by the
`tenant-service` once the cluster tokens have been retrieved and
decoded.


fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-541-1`

PR #539, opened 2018-02-27
Add github.com/pilu/fresh to glide.yaml 

stooke:
The Makefile requires 'fresh', but it isn't automatically captured by the Glide tool.
Adding fresh and it's dependencies ensures a clean checkout and 'make dev' should compile.

PR #543, opened 2018-02-28
Add logging URL to tenant attributes

stooke:
This PR adds the logging URL (to accompany the console and metrics URLs) to the tenant attributes.

After this is merged, the logging URL will be retrieved from the Auth service, and added to the endpoint /api/tenant.

Following this PR, a PR for the WIT service will consume the logging URL.


glide.yaml
```diff
@@ -67,6 +69,9 @@ import:
 - package: github.com/bitly/go-simplejson
   version: ^0.5.0
 - package: github.com/google/go-querystring
+- package: github.com/pilu/config
```
aslakknutsen: technically not part of this PR

stooke: Absolutely true.  That's from PR #539, but I can't build tenant without it.

aslakknutsen:
[test]

token/service.go
```diff
@@ -57,10 +57,7 @@ func (s *tokenService) ResolveTargetToken(ctx context.Context, target, token str
 	if err != nil {
 		return "", "", errors.Wrapf(err, "error while decoding the token for %s", target)
 	}
-	if externalToken.Username == nil {
```
aslakknutsen: ?

stooke: The Auth service changed its API contract to require Username on output.  Because of this, Goa changed the generated type from a pointer-to-string to an actual string.  Because it's an actual string (so never nil) and required (so never length 0) I took out the test and removed a later indirection.

If you wish I can add a test to ensure non-empty string; I just felt that was redundant.

aslakknutsen: Re-enable the check for empty string. We want to know specifically if it's "" instead of some random unknown failure later. Can't trust @alexeykazakov ;)

aslakknutsen: Unless @alexeykazakov can confirm it's always there now. :)

alexeykazakov: It's always there unless we have a bug in Auth :) So, don't trust me :)

alexeykazakov: To be clear. Username should be always there. It's why we changed it from optional to required. But trust no one! Auth can have bugs (hard to believe though). Check the output and report an error if it's not what you expected.

stooke: Test added.  Best part is the previous code had a bug in it: missing %s !

fabric8cd:
@stooke snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-543-4`

PR #546, opened 2018-03-03
Add a retry loop to validate ResourceQuotas has been applied

aslakknutsen:
ResourceQuota resource has a async nature so attempting to fill the quota
before the quota has been applied cause 403 Forbidden errors.

Enter a retry get loop after create of ResourceQuota to wait for it
to finish.

Related to openshiftio/openshift.io#2381

### DIFF
--- a/controller/tenant.go
+++ b/controller/tenant.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


fabric8cd:
@aslakknutsen snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-546-1`

PR #579, opened 2018-04-27
API to delete/deprovision a user from Tenant (OSIO#3338)

xcoulon:
Refactored the `tenants` controller to include a new `openshift`
service that provides a method to delete a namespace using the
OSO API. This service can use a different HTTP client transport,
which allows for testing with the `go-vcr` library.

If a namespace deletion fails, the whole deletion stops, except for
`403 Forbidden` responses from OSO which can be returned if the
namespace to delete does not exist. In that case, there's no reason
to halt the tenant deletion.

Fixes openshiftio/openshift.io#3338

### DIFF
--- a/controller/tenants.go
+++ b/controller/tenants.go
@@ -1,28 +1,49 @@
 package controller
 
 import (
+	"fmt"
+	"reflect"
+
 	"github.com/fabric8-services/fabric8-tenant/app"
 	"github.com/fabric8-services/fabric8-tenant/cluster"
 	"github.com/fabric8-services/fabric8-tenant/jsonapi"
+	"github.com/fabric8-services/fabric8-tenant/openshift"
 	"github.com/fabric8-services/fabric8-tenant/tenant"
 	"github.com/fabric8-services/fabric8-tenant/token"
+	"github.com/fabric8-services/fabric8-tenant/user"
 	"github.com/fabric8-services/fabric8-wit/errors"
+	"github.com/fabric8-services/fabric8-wit/log"
 	"github.com/goadesign/goa"
 )
 
 // TenantsController implements the tenants resource.
 type TenantsController struct {
 	*goa.Controller
-	tenantService  tenant.Service
-	resolveCluster cluster.Resolve
+	tenantService          tenant.Service
+	resolveTenant          tenant.Resolve
+	userService            user.Service
+	openshiftService       openshift.Service
+	resolveCluster         cluster.Resolve
+	defaultOpenshiftConfig openshift.Config
 }
 
 // NewTenantsController creates a tenants controller.
-func NewTenantsController(service *goa.Service, tenantService tenant.Service, resolveCluster cluster.Resolve) *TenantsController {
+func NewTenantsController(service *goa.Service,
+	tenantService tenant.Service,
+	userService user.Service,
+	openshiftService openshift.Service,
+	resolveTenant tenant.Resolve,
+	resolveCluster cluster.Resolve,
+	defaultOpenshiftConfig openshift.Config,
+) *TenantsController {
 	return &TenantsController{
-		Controller:     service.NewController("TenantsController"),
-		tenantService:  tenantService,
-		resolveCluster: resolveCluster,
+		Controller:             service.NewController("TenantsController"),
+		tenantService:          tenantService,
+		resolveTenant:          resolveTenant,
+		userService:            userService,
+		openshiftService:       openshiftService,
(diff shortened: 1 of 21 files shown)


controller/tenants.go
```diff
@@ -73,3 +94,57 @@ func (c *TenantsController) Search(ctx *app.SearchTenantsContext) error {
...
 }
+
+// Delete runs the `delete` action to deprovision a user
+func (c *TenantsController) Delete(ctx *app.DeleteTenantsContext) error {
+	if !token.IsSpecificServiceAccount(ctx, "fabric8-auth") {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("Wrong token"))
+	}
+	tenantID := ctx.TenantID
+	// fetch the cluster the user belongs to
+	usr, err := c.userService.GetUser(ctx, tenantID)
```
aslakknutsen: Not really needed. We have that from our own DB

openshift/service.go
```diff
@@ -0,0 +1,67 @@
...
+		return errors.Wrapf(err, "unable to delete the namespace from the API endpoint")
+	}
+	body := buf.Bytes()
+	// only report error if the operation did not return a 2xx (OK) or 403 (Forbidden)
+	// actually, Openshift checks if the namespace belongs to the user even before checking if it exists...
+	if resp.StatusCode == http.StatusForbidden {
+		// let's log the error, nonetheless
+		log.Warn(ctx, map[string]interface{}{"namespace": namespace, "message": body}, "failed to delete namespace (but it probably does not exist)")
+	}
+	if resp.StatusCode >= 400 && resp.StatusCode != http.StatusForbidden {
```
aslakknutsen: else if?

controller/tenants.go
xcoulon: ok, fixed in 764fc32af4f0ecb16553db2bfe4b5a9101d018d6

controller/tenants.go
```diff
@@ -73,3 +93,53 @@ func (c *TenantsController) Search(ctx *app.SearchTenantsContext) error {
...
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"err":         err,
+				"cluster_url": namespace.MasterURL,
+				"tenant_id":   tenantID,
+			}, "unable to fetch cluster for user")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+		}
+
+		// openshiftConfig := openshift.NewConfig(c.defaultOpenshiftConfig, usr, clustr.User, clustr.Token, clustr.APIURL)
```
surajssd: comment, needs removal ?

design/tenant.go
```diff
@@ -170,6 +170,22 @@ var _ = a.Resource("tenants", func() {
 		a.Response(d.Unauthorized, JSONAPIErrors)
 	})
 
+	a.Action("delete", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.GET("/:tenantID"),
```
alexeykazakov: tenantID is an identity/user ID, right?

xcoulon: yes, it's called `tenant` in this service

controller/tenants.go
xcoulon: right, let me do that

controller/tenants.go
```diff
@@ -73,3 +93,53 @@ func (c *TenantsController) Search(ctx *app.SearchTenantsContext) error {
...
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("Wrong token"))
+	}
+	tenantID := ctx.TenantID
+	namespaces, err := c.tenantService.GetNamespaces(tenantID)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+	for _, namespace := range namespaces {
+		// fetch the cluster info
+		clustr, err := c.resolveCluster(ctx, namespace.MasterURL)
```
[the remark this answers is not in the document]
xcoulon: no, as @aslakknutsen pointed out, namespaces could be located on different clusters, so it's safer to do it like that

fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-579-3`

fabric8cd:
@xcoulon snapshot fabric8-tenant image is available for testing.  `docker pull docker.io/fabric8/fabric8-tenant:SNAPSHOT-PR-579-4`

PR #632, opened 2018-09-11
chore: refactoring of the components that comunicate with auth service

MatousJobanek:
This is the first step of a complete rewrite of the tenant service and is a preparation for OSIO on OSD support.

This PR contains a code that was separated (and slightly modified to fit here) from the whole rewrite that can be found here: https://github.com/MatousJobanek/fabric8-tenant/tree/v2 (FYI the code in the branch is still work-in-progress)

#### Proposed changes 
* It "only" moves the code and integrates together those parts that are somehow related to auth service. 
  * Everything that is related to getting information about users and token management is moved to `auth` package and the main entry point for the communication (and auth client management) is [auth-service](https://github.com/fabric8-services/fabric8-tenant/compare/master...MatousJobanek:refactoring?expand=1#diff-ec041ec477251a53b62fb402b364d73bR19).
  * Everything that is related to clusters is located in `cluster` package and the main entry point is
[cluster-service](https://github.com/fabric8-services/fabric8-tenant/compare/master...MatousJobanek:refactoring?expand=1#diff-c0b4d0964c7a7f269081aa6229c3ed30R34).
* Previously, there was used a combination of resolve functions that were integrated with each other and were internally used. Instead of it, I'm proposing a usage of service objects and passing the instances to the methods/functions.
* As part of this refactoring I was trying to use the fabric8-common repository when possible.
* Tests are technically almost the same as in the previous version, there was just slightly changed the setup of the tests.


auth/publickey.go
```diff
@@ -1,29 +1,27 @@
...
 	"github.com/fabric8-services/fabric8-wit/rest"
 	errs "github.com/pkg/errors"
-	jose "gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2"
 )
 
 // GetPublicKeys returns the known public keys used to sign tokens from the auth service
-func GetPublicKeys(ctx context.Context, authURL string, options ...configuration.HTTPClientOption) ([]*rsa.PublicKey, error) {
-	client, err := auth.NewClient(authURL, "", options...) // no need for a token when calling this endpoint
+func (s *Service) GetPublicKeys() ([]*rsa.PublicKey, error) {
```
alexeykazakov: Any particular reason to keep this method in a separate file (from auth/service.go)?

auth/service.go
```diff
@@ -0,0 +1,233 @@
+package auth
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/dgrijalva/jwt-go"
+	commonConfig "github.com/fabric8-services/fabric8-common/configuration"
```
alexeykazakov: Package names should be lower case.

auth/service.go
```diff
@@ -0,0 +1,233 @@
...
+	commonErrors "github.com/fabric8-services/fabric8-common/errors"
```
alexeykazakov: Package names should be lower case.

auth/service.go
```diff
@@ -0,0 +1,233 @@
...
+			&goaclient.JWTSigner{
+				TokenSource: &goaclient.StaticTokenSource{
+					StaticToken: &goaclient.StaticToken{
+						Value: token,
+						Type:  "Bearer"}}})
+	}
+	return client, nil
+}
+
+func (s *Service) getOAuthToken(ctx context.Context) (*string, error) {
```
alexeykazakov: This method is a good candidate for fabric8-common. Many services need to load Service Account tokens from Auth.

auth/service.go
```diff
@@ -0,0 +1,233 @@
...
+	return s.newClient(s.SaToken)
+}
+
+func (s *Service) newClient(token string) (*authclient.Client, error) {
+	client, err := newClient(s.Config.GetAuthURL(), token, s.ClientOptions...)
+	if err != nil {
+		return nil, err
+	}
+	if token != "" {
+		client.SetJWTSigner(
```
alexeykazakov: If we use Auth client only when we start the service to load keys and SA token (we will need to change it later to support dynamic keys and SA token loading) then it's fine to use JWTSigner only.
But if we also use Auth during handing some tenant requests then we also would need to forward request ID header. But we would need the original context for that.

auth/service.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := c.ShowUsers(ctx, authclient.ShowUsersPath(subject(userToken)), nil, nil)
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while doing the request")
+	}
+	defer res.Body.Close()
```
alexeykazakov: Also read the body in defer.

openshift/whoami_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/fabric8-services/fabric8-tenant/configuration"
+	commonConfig "github.com/fabric8-services/fabric8-common/configuration"
```
alexeykazakov: Package name.

auth/auth_client.go
```diff
@@ -8,18 +8,14 @@ import (
 
 	"github.com/fabric8-services/fabric8-auth/errors"
 
+	commonConfig "github.com/fabric8-services/fabric8-common/configuration"
```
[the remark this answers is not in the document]
MatousJobanek: ah, I forgot about this convention - thanks for pointing it out.
> Why not just "github.com/fabric8-services/fabric8-common/configuration" (== configuration)?

because I wanted to distinguish the tenant config and common config

I'll change it to `cfg`

auth/publickey.go
MatousJobanek: well, the `auth/service.go` file seemed to be big enough and originally it was in this file `auth/publickey.go` so I kept it there. But you are right that it would make more sense to keep all methods with the same receiver in the same file :+1: and there is no significant reason for keeping it separated.

auth/auth_client.go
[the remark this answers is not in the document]
MatousJobanek: fixed in 1a94b7f

auth/publickey.go
MatousJobanek: refactored in d84fbd8

auth/service.go
MatousJobanek: fixed in 1a94b7f

auth/service.go
MatousJobanek: fixed in 1a94b7f

auth/service.go
MatousJobanek: added in 60adab6

openshift/whoami_test.go
MatousJobanek: fixed in 1a94b7f

auth/service.go
MatousJobanek: I'm afraid that this won't be easy to do (if possible at all) - it uses goa generated objects from the auth design which then leads to conflicts in imports. An example is the auth `Client` as well as the `TokenExchange` - both of them are generated and to be able to use it in the code I need to [generate](https://github.com/fabric8-services/fabric8-tenant/blob/master/Makefile#L218) them directly in the projects. This leads to the situation where I have a method `GetOAuthToken` expecting a `Client` or `TokenExchange` from the generated package located in common repo (eg: `github.com/fabric8-services/fabric8-common/auth"`), but in the tenant service I have these objects generated in a different package: `github.com/fabric8-services/fabric8-tenant/auth/client` so I cannot pass it to the function. 
Is there some way to work around this that I wouldn't be aware of?

auth/service.go
```diff
@@ -0,0 +1,331 @@
...
+	"encoding/json"
+	"github.com/fabric8-services/fabric8-wit/rest"
+	errs "github.com/pkg/errors"
+)
+
+type Service struct {
+	Config        *configuration.Data
+	ClientOptions []commonconf.HTTPClientOption
+	SaToken       string
+}
```
nurali-techie: is adding interface auth_service is useful, like we have for cluster_service ?

auth/service_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	t.Run("empty access token", func(t *testing.T) {
+		// when
+		_, _, err := authService.ResolveUserToken(context.Background(), "some_valid_openshift_resource", "")
+		// then
+		require.Error(t, err)
+	})
+}
+
+func TestResolveServiceAccountToken(t *testing.T) {
```
nurali-techie: sub-tests are not calling `ResolveSaToken()` but calling `ResolveTargetToken()` which is anyway sort of cover by `TestResolveUserToken()` test.  why this `TestResolveServiceAccountToken()` test is needed OR can we have test for `ResolveSaToken()` .. at current state of test, I see it's almost duplicate and not adding much value.

MatousJobanek: fixed in 161429

auth/service.go
MatousJobanek: For now, it wouldn't bring any additional value as there is no additional implementation of the `auth.Service` interface - that is also why I didn't add it there. 
In the future, I have a plan to add another dummy impl to support local testing without running auth client, then having the interface there would make sense.
As for the `cluster.Service`, yeah, I could have removed it to make the code consistent,I just didn't see it as a big deal :-). 
I can do anything of these options:
* remove interface from cluster
* add interface to auth
* keep it as it is

I don't have strong opinion on it ;-)

nurali-techie: I am fine with - keep it as it is :+1:

cluster/service.go
```diff
@@ -51,74 +45,78 @@ type Stats struct {
...
+		return fmt.Errorf("failed to load the list of clusters during service initialization: %s", err)
 	}
 	go func() {
-		for range cacheRefresher.C { // while the `cacheRefresh` ticker is running
+		for range s.cacheRefresher.C { // while the `cacheRefresh` ticker is running
 			s.refreshCache(context.Background())
 		}
 	}()
-	return s, nil
+	return nil
```
nurali-techie: I don't see any benefit of adding separate `Start()` rather it add responsibility on consumer to call `Start()` without fail.  any specific reason for adding `Start()` ??

Also currently we are failing `NewClusterService()` if `refreshCache()` failed .. this is wrong thing .. like `refreshCache()` failed due to temporary network issue / service unavailable issue.

I guess, we can keep `NewClusterService()` as it is (without adding `Start()`) and not to fail when `refreshCache()` return error and hope that future call to `refreshCache()` will populate some cluster data. Cc: @alexeykazakov WDYT?

test/doubles/test_doubles.go
```diff
@@ -0,0 +1,37 @@
...
+		if err != nil {
+			return nil, r, err
+		}
+		options = append(options, commonConfig.WithRoundTripper(r))
+	}
+	config, err := LoadTestConfig()
+	if err != nil {
+		return nil, r, err
+	}
+	config.Set(configuration.VarAuthURL, authURL)
```
nurali-techie:
```
os.Setenv("F8_AUTH_URL", "http://auth.test.com") // check F8_ convention
config.GetAuthURL() // this will return http://auth.test.com
```
so you no need to export `varAuthURL` and also other variable you have exported.

controller/tenant.go
```diff
@@ -68,42 +65,32 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
 	}
 
 	// fetch the cluster the user belongs to
-	user, err := c.userService.GetUser(ctx, ttoken.Subject())
+	user, err := c.authClientService.NewUser(ctx)
 	if err != nil {
```
nurali-techie: this is `GetUser()` rather `NewUser()`
and I guess, taking ctx and token as input argument would be better.  Input to method clarify what all it depends on.  In go ctx is such a object where you can add param and pass only ctx around but that's not right.

cluster/service.go
```diff
@@ -51,74 +45,78 @@ type Stats struct {
...
-	cachedClusters   []Cluster
-	decode           token.Decode
-	clientOptions    []configuration.HTTPClientOption
+func (s *clusterService) GetCluster(ctx context.Context, target string) (*Cluster, error) {
+	for _, cluster := range s.GetClusters(ctx) {
+		if cleanURL(target) == cleanURL(cluster.APIURL) {
+			return &cluster, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to resolve cluster")
```
nurali-techie: Here you should return copy of Cluster rather actual Cluster (`*Cluster`), similar to GetClusters().  This is imp as this object are kept in cache and we can't pass reference around.

cluster/service.go
```diff
@@ -30,17 +28,13 @@ type Cluster struct {
...
-	return url
-}
+type GetCluster func(ctx context.Context, target string) (*Cluster, error)
 
 // Service the interface for the cluster service
 type Service interface {
-	GetClusters(context.Context) ([]Cluster, error)
-	Stats() Stats
+	GetCluster(ctx context.Context, target string) (*Cluster, error)
+	GetClusters(ctx context.Context) []Cluster
```
nurali-techie: for API consistency, you can have GetCluser() return `Cluster` rather `*Cluster`. (nice to have comment)

main.go
```diff
@@ -94,7 +92,14 @@ func main() {
 	templateVars["KEYCLOAK_OSO_ENDPOINT"] = keycloakConfig.CustomBrokerTokenURL("openshift-v3")
 	templateVars["KEYCLOAK_GITHUB_ENDPOINT"] = fmt.Sprintf("%s%s?for=https://github.com", config.GetAuthURL(), authclient.RetrieveTokenPath())
 
-	publicKeys, err := token.GetPublicKeys(context.Background(), config.GetAuthURL())
+	authService, err := auth.NewAuthService(config)
+	if err != nil {
+		log.Panic(nil, map[string]interface{}{
+			"err": err,
+		}, "failed to fetch service account token")
```
nurali-techie: something wrong with error msg ;-)

cluster/service.go
MatousJobanek: > I don't see any benefit of adding separate Start() rather it add responsibility on consumer to call Start() without fail. any specific reason for adding Start()

there was a reason in the new implementation I took the code from - I was using it for local test purposes. Technically, I could work around it by setting a big refresh interval. So if you think that it doesn't make sense, then I can change it back to the original implementation and if I still need the `start()` method when I incorporate also the changes of the core logic, then I'll bring it back.

> Also currently we are failing NewClusterService() if refreshCache() failed .. this is wrong thing .. like refreshCache() failed due to temporary network issue / service unavailable issue.

this is what I was thinking about as well - if it should fail or not, but I kept the original implementation. Anyway, we need to have the tenant service ready once it's started, so ignoring the error is not the best idea I think. We can incorporate a [retry](https://github.com/MatousJobanek/fabric8-tenant/blob/v2/retry/retry.go) function that would try to get the list of clusters several times with some delay between the attempts. that would solve temporary network issues

test/doubles/test_doubles.go
MatousJobanek: I'm just quite afraid of setting the environment variables because I experienced a lot of problems with it in previous projects - mainly when it is possible to do it programmatically.

controller/tenant.go
MatousJobanek: > this is GetUser() rather NewUser()

yeah, I wasn't sure about this name. it can be `GetUser()` because it gets info about user from the auth sevice, but can be `NewUser()` because it constructs a new instance of `auth.User` where the params are taken from more sources.

> and I guess, taking ctx and token as input argument would be better. Input to method clarify what all it depends on. In go ctx is such a object where you can add param and pass only ctx around but that's not right.

well, it depends on the token that is bundled inside of the context of an incoming request. The original code was just doing the same thing (retrieving token from context) on several places when it can be moved directly to the method and can be hidden as this is just an implementation detail. I know that the code of retrieving the token information left in the original places, but will be removed in the future.

I wrote the function with the idea "Create me an instance of the `auth.User` for the given context, where the user represents the client who sent the request"

cluster/service.go
MatousJobanek: changed in 84fe6dc

main.go
MatousJobanek: ??? I'm probably blind...

nurali-techie: in case `NewAuthService()` return error, we should logging relevant error_msg and not "failed to fetch service account token". :-)

cluster/service.go
```diff
@@ -51,74 +45,78 @@ type Stats struct {
...
 	err := s.refreshCache(context.Background())
 	if err != nil {
-		log.Error(nil, map[string]interface{}{"error_message": err}, "failed to load the list of clusters during service initialization")
-		return nil, err
+		return fmt.Errorf("failed to load the list of clusters during service initialization: %s", err)
 	}
 	go func() {
-		for range cacheRefresher.C { // while the `cacheRefresh` ticker is running
+		for range s.cacheRefresher.C { // while the `cacheRefresh` ticker is running
 			s.refreshCache(context.Background())
```
alexeykazakov: Log the error here. Don't just ignore it.

cluster/service.go
alexeykazakov: 1. We need clusters to be loaded when starting a tenant service. Without having it the service is not functional. So, we the tenant service readiness prob should fail if we can't load clusters during startup. OpenShift will retry to re-start the pod a few times before giving up. If it's a temporal network issue then it will help. If not, then the old tenant service version with old cluster cache will continue to work. So no outages. Good :)

2. I don't see any issues with the new Start() method. NewClusterService() won't fail if cluster service is not available because it's not loading clusters but we call Start() in main too so the tenant service startup will fail on Start() which is good (see # 1 above)

3. Go routine started by Start() will keep working even if some attempts to refresh the cache fail. The service will be using the old cache until the network issue is gone. We just need to log the error here (you missed it and I've added a comment for that).

So, to summarize:
- Tenant service startup should fail if we can't load the initial list of clusters.
- If cache refresh fails then we need to continue our attempts to refresh it and keep using the old cache meanwhile.
It looks like this PR covers all of these above so we are good (just log the error when refresh fails).

test/doubles/test_doubles.go
alexeykazakov: While it's OK to use config.Set() in tests (if you don't some tests in parallel which share the same viper instance. Viper is not thread safe, so, you can't use Set() if the viper instance is used by multiple go routines). But I also don't like exporting var names without a reason (it's not that critical though.. it's more about code cleanness). It's really an implementation detail and configuration consumers don't have to know about that.

When you need to get some custom configuration for you test then you can:

1. config.Set("auth.url", authURL) - it's fine to hardcode the var name in tests. If you change the var name in configuration then this test will fail. So, there is no risk of doing it. And I like this more then exporting the var names.

2. Or you can set the env var as @nurali-techie suggested and reload the config. It's safe if you take care of re-setting it back to the original value. See an example here - https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration_blackbox_test.go#L55-L61
It can be quite verbose though..

controller/tenant.go
alexeykazakov: It's definitely GetUser() not NewUser() :) You are calling an auth service to get the existing user. Not to create a new User in the Auth service. The fact that it's creating a new instance from auth.User struct is just an implementation detail of the service. 

Regarding context and token... I agree with @MatousJobanek. We just pass the incoming contest to the service. Extracting the token (which was added there by goa JWT middleware) and creating a new request for Auth is an implementation detail of the service.

cluster/service.go
MatousJobanek: blindly copied from the original code :-).
fixed in 5d94d71

cluster/service.go
```diff
@@ -51,74 +45,83 @@ type Stats struct {
...
-	if err != nil {
-		return err
-	}
-	client.SetJWTSigner(
-		&goaclient.JWTSigner{
-			TokenSource: &goaclient.StaticTokenSource{
-				StaticToken: &goaclient.StaticToken{
-					Value: s.serviceToken,
-					Type:  "Bearer"}}})
+	client, err := s.authService.NewSaClient()
```
golangcibot: ineffectual assignment to `err`

main.go
MatousJobanek: changed in 65ea36f

controller/tenant.go
MatousJobanek: renamed in 2547d6c

cluster/service.go
MatousJobanek: fixed in 14c8dba

test/doubles/test_doubles.go
MatousJobanek: > Viper is not thread safe

I didn't know that. Anyway, setting system environments is not thread-safe either, is it?

> But I also don't like exporting var names without a reason

ok, make sense - fixed in 3101416 

>  It's safe if you take care of re-setting it back to the original value....
It can be quite verbose though.

yes, the verbosity is the reason why I didn't want to go this way. Currently, it is used in 6 test cases, which means that I would have to copy-paste the same code in all places (technically, I could use test suites or subtests). In addition, as I'm trying to add more this kind of integration tests, where having config initialized is necessary, then the code duplication will be much bigger. But that is only my point of view ;-). I'm open to any solution

cluster/service.go
alexeykazakov: Still not fixed :)

test/doubles/test_doubles.go
alexeykazakov: In this case using config.Set("auth.url", authURL) is the best solution IMO (short and simple). But sometimes you need to test the configuration itself. Then you have to use env vars to test how configuration is loaded.

cluster/service.go
```diff
@@ -51,74 +45,81 @@ type Stats struct {
...
+	return Cluster{}, fmt.Errorf("unable to resolve cluster")
 }
 
-func (s *clusterService) GetClusters(ctx context.Context) ([]Cluster, error) {
+func (s *clusterService) GetClusters(ctx context.Context) []Cluster {
 	s.cacheRefreshLock.RLock()
 	log.Debug(ctx, nil, "read lock acquired")
 	clusters := make([]Cluster, len(s.cachedClusters))
 	copy(clusters, s.cachedClusters)
 	s.cacheRefreshLock.RUnlock()
```
alexeykazakov: Always unlock in defer right after locking, so if there is panic you won't block the cache forever:

s.cacheRefreshLock.RLock()
defer s.cacheRefreshLock.RUnlock()
clusters := make([]Cluster, len(s.cachedClusters))
copy(clusters, s.cachedClusters)
...

cluster/service.go
```diff
@@ -172,11 +173,3 @@ func (s *clusterService) refreshCache(ctx context.Context) error {
 	log.Debug(ctx, nil, "write lock released")
```
alexeykazakov: Always unlock in defer right after locking, so if there is panic you won't block the cache forever.

cluster/service.go
MatousJobanek: what the hell? for some reason, this [commit](14c8dba) removed the part that I added before... probably result of some crazy rebasing/stashing...
fixed again in: 008645a 
sorry for that and thanks for finding it

cluster/service.go
MatousJobanek: done in 2252176

cluster/service.go
MatousJobanek: done in 2252176

test/doubles/test_doubles.go
```diff
@@ -0,0 +1,63 @@
...
+	config, err := configuration.GetData()
+	require.NoError(t, err)
+
+	authService := &auth.Service{
+		Config:        config,
+		ClientOptions: clientOptions,
+	}
+	return authService, r, func() {
+		if r != nil {
+			r.Stop()
```
golangcibot: Error return value of `r.Stop` is not checked

cluster/service_test.go
```diff
@@ -8,46 +8,40 @@ import (
...
 	// given
-	r, err := recorder.New("../test/data/cluster/resolve_cluster.fast", recorder.WithJWTMatcher())
-	require.NoError(t, err)
-	defer r.Stop()
-	authURL := "http://fast.authservice"
-	resolveToken := token.NewResolve(authURL, configuration.WithRoundTripper(r))
+	reset := testdoubles.SetEnvironments(testdoubles.Env("F8_AUTH_TOKEN_KEY", "foo"))
+	defer reset()
+	authService, cleanup := testdoubles.NewAuthService(t, "../test/data/cluster/resolve_cluster.fast", "http://fast.authservice", recorder.WithJWTMatcher)
+	defer cleanup()
```
golangcibot: Error return value of `cleanup` is not checked

cluster/service_test.go
```diff
@@ -89,35 +80,28 @@ func TestResolveCluster(t *testing.T) {
...
 	// given
-	r, err := recorder.New("../test/data/cluster/resolve_cluster.slow", recorder.WithJWTMatcher())
-	require.NoError(t, err)
-	defer r.Stop()
-	authURL := "http://slow.authservice"
-	resolveToken := token.NewResolve(authURL, configuration.WithRoundTripper(r))
+	reset := testdoubles.SetEnvironments(testdoubles.Env("F8_AUTH_TOKEN_KEY", "foo"))
+	defer reset()
+	authService, cleanup := testdoubles.NewAuthService(t, "../test/data/cluster/resolve_cluster.slow", "http://slow.authservice", recorder.WithJWTMatcher)
+	defer cleanup()
```
golangcibot: Error return value of `cleanup` is not checked

cluster/service_test.go
MatousJobanek: fixed cb77388

cluster/service_test.go
MatousJobanek: fixed cb77388

auth/service_test.go
```diff
@@ -0,0 +1,192 @@
...
+		require.NoError(t, err)
+		assert.Equal(t, "user_foo", username)
+		assert.Equal(t, "an_openshift_token", accessToken)
+	})
+
+	t.Run("invalid resource", func(t *testing.T) {
+		// when
+		_, _, err := authService.ResolveUserToken(context.Background(), "some_invalid_resource", tok.Raw)
+		// then
+		require.Error(t, err)
```
alexeykazakov: Check error message too to make sure you got the expected error. Here and in other places too.
Maybe we can benefit from a function like this https://github.com/fabric8-services/fabric8-auth/blob/master/test/error.go#L11
Usage: https://github.com/fabric8-services/fabric8-auth/blob/4a020fbf23eff0b479c6a98d2a8f8452ab6f2c57/authorization/space/service/space_service_blackbox_test.go#L28

auth/service_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	testsupport "github.com/fabric8-services/fabric8-tenant/test"
+	"github.com/fabric8-services/fabric8-tenant/test/doubles"
+	"github.com/fabric8-services/fabric8-tenant/test/recorder"
+	goajwt "github.com/goadesign/goa/middleware/security/jwt"
+)
+
+func TestResolveUserToken(t *testing.T) {
```
alexeykazakov: Can you improve AuthService coverage please?

cluster/service.go
nurali-techie: the changes related to `refreshCache()` [link](https://github.com/fabric8-services/fabric8-tenant/commit/14c8dba35390dd43c5f6c41a48fea7eccb9aa6a8#diff-c0b4d0964c7a7f269081aa6229c3ed30R78).
I think, we should log error the way you were doing before.  It gives idea that `refreshCache()` is failing and there will be retry.

MatousJobanek: see this comment https://github.com/fabric8-services/fabric8-tenant/pull/632#discussion_r219769406

auth/service_test.go
nurali-techie: agree with @MatousJobanek , I can see that most of the code is covered [link](https://codecov.io/gh/fabric8-services/fabric8-tenant/pull/632/diff) and coverage looks good for auth_service :-) :+1:

alexeykazakov: Maybe it was a codecov glitch but the report showed some big parts of the service not covered. Or maybe it's your latest commit with SA token tests... Anyway it's much better now!

PR #633, opened 2018-09-11
feat: add sentry integration

MatousJobanek:
All internal errors will be sent to Sentry

* for the integration the code from fabric8-common repo was used
* to use the same source of the logic, all other imports of the logger was changed from `fabric8-wit` to `fabric8-common`

https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2305


MatousJobanek:
I tested it locally, but with a slightly different setup and with very basic code - I didn't check it in the whole flow. That is probably why it worked in my case. My bad :-(.

MatousJobanek:
Btw, my original version was similar to what we have in [ike-prow-plugins](https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/plugin_init.go#L126) project. The sentry integration is tied to the logrus and the record is sent whenever a panic, fatal or error log is created. That ensures that all problems are reported to sentry, the drawback of that solution is that the number of records sent to sentry can be a bit higher than needed - in the case when there are reported several error logs (in different functions) for one failure/problem.

alexeykazakov:
Let's discuss this sentry/token manager thing next week.

alexeykazakov:
@MatousJobanek I've created a PR to fix the sentry issue with missing token manager/user info in fabric8-common - https://github.com/fabric8-services/fabric8-common/pull/11

main.go
```diff
@@ -259,13 +260,41 @@ func migrate(db *gorm.DB) {
...
+}
+
+func extractUserInfo(userService user.Service) func(ctx context.Context) (*raven.User, error) {
+	return func(ctx context.Context) (*raven.User, error) {
+		userToken := goajwt.ContextJWT(ctx)
+		if userToken == nil {
+			return nil, fmt.Errorf("no token found in context")
+		}
+		ttoken := &controller.TenantToken{Token: userToken}
+		user, err := userService.GetUser(ctx, ttoken.Subject())
```
alexeykazakov: You don't have to call Auth service to get User Info. You can get it from the token.
userID := ttoken.Subject()
username := ttoken.Username()
email := ttoken.Email()

MatousJobanek:
> Do you really want to report all errors to sentry?

these errors occur in a goroutine, if there is a problem with creating/updating/deleting a namespace, then these errors are logged to the console and the error will never get to this [point](https://github.com/fabric8-services/fabric8-tenant/pull/633/files#diff-b160c561ec25cc415f1cdbe6d0bebb8dR171) to be reported to sentry. If there is such a problem, then it usually means that any of the user's namespaces can be in a bad state. 
Then, to figure out what is the problem, I have to go to the console and try to find the correct report of the error. From my point of view, it would be better having them directly in the sentry. But that was just my idea/feeling after my latest experience when I was trying to figure out why there was missing one namespace in a test account - that is the reason why I introduced these calls.
Anyway, you are right that it would be good to put more info/params there.

WDYT @alexeykazakov , does it make sense - having these calls to sentry or not?

main.go
MatousJobanek: I didn't realize this - I sometimes get lost in these auth/token things. thanks for pointing it out :+1: 
fixed in 4d77efe

sentry/sentry.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return *pointer
+}
+
+// LogError logs the given error and reports it to sentry
+func LogError(ctx context.Context, fields map[string]interface{}, err error, message string) {
+	sentryError := fmt.Errorf("an error occured with a message: \n%s\n with fields: \n%s\n and caused by: \n%s", message, fields, err)
+	sentry.Sentry().CaptureError(ctx, sentryError)
+
+	fields["error"] = err
```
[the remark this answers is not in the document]
MatousJobanek: I wasn't sure which one I should use - in the previous logic, there were used both versions.
changed in ced284b

alexeykazakov:
@MatousJobanek OK... I think we can add them and see how it goes. If it helps then great. If it doesn't we can always remove it or improve the reporting (even better) :)

MatousJobanek:
> But could you please add tests for changed code?

I tried my best aa1da93
while writing it I realized that the retrieval of params from token could be more robust :-)

configuration/configuration.go
```diff
@@ -143,6 +146,10 @@ func (c *Data) setConfigDefaults() {
 	c.v.SetDefault(varClientSecret, "tenantsecretNew")
 }
 
+func (c *Data) Set(key string, value interface{}) {
+	c.v.Set(key, value)
```
alexeykazakov: Oh.. I actually missed it in the other PR... So, you need to export Set() method in Configuration to be able to use it in tests. But it's error prone because Set() can be used in tests only. It may cause issues if used in runtime code because Set() is not thread safe in viper...

So, you have to use env vars :( or as alternative you could use some configuration interfaces instead of passing the whole configuration.Data.
For example something like this: https://github.com/fabric8-services/fabric8-auth/blob/master/account/service/tenant.go#L26
https://github.com/fabric8-services/fabric8-auth/blob/master/account/service/tenant.go#L36
https://github.com/fabric8-services/fabric8-auth/blob/master/account/service/tenant_whitebox_test.go#L37

MatousJobanek: solved in 5136ba5
the same approach (the same helper functions) I'm [using](cb77388) in the other PR, so merging should be easier.

sentry/sentry_test.go
```diff
@@ -0,0 +1,84 @@
...
+		assert.Empty(t, "", user.Username)
+		assert.Empty(t, "", user.Email)
+	})
+
+	t.Run("context without token", func(t *testing.T) {
+		// when
+		_, err := extractUserInfo()(context.Background())
+
+		// then
+		require.Error(t, err)
```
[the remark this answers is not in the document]
MatousJobanek: good idea. the tests are improved here fd0e85f

test/doubles/test_doubles.go
```diff
@@ -0,0 +1,25 @@
...
+func Env(key, value string) Environment {
+	return Environment{key: key, value: value}
+}
+
+type Environment struct {
+	key, value string
+}
+
+func SetEnvironments(environments ...Environment) func() {
+	originalValues := make([]Environment, len(environments))
```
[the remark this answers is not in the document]
MatousJobanek: my fingers were too fast - I wanted to write:
```go
originalValues := make([]Environment, 0, len(environments))
```
at least this is the format I'm used to - using a slice with the predefined capacity avoids reallocating of the underlying array in the append function
Sorry for this stupid mistake.

Issue #638, opened 2018-09-26
Reset environment failed


kishansagathiya:
Resetting environment failed
Request
```
Request URL: https://prod-preview.openshift.io/api/user/services
Request Method: DELETE
Status Code: 500 Internal Server Error
```
Error
```
{
  "errors": [
    {
      "code": "internal_error",
[...]

PR #639, opened 2018-09-26
fix: replace conversion to string by fmt.Sprint to avoid panics

MatousJobanek:
replace conversion to string by `fmt.Sprint` to avoid panics when the value is nil

fixes: #638


aslakknutsen:
@MatousJobanek Know anything about how toggles will handle a "null" ? Might cause some 'i'll set anything to true' oddy behavior ?

MatousJobanek:
It should not do any odd behavior, but good point that setting "<nil>" is not the best idea. 
this should be more safe 3dd4b62

PR #634, opened 2018-09-17
feat: simplify template management

MatousJobanek:
This PR simplifies the management of the templates that are used for operations over tenant namespaces (provisions/delete/update).
The location of the templates is changed from the external projects ([che](https://github.com/fabric8-services/fabric8-tenant-che), [jenkins](https://github.com/fabric8-services/fabric8-tenant-jenkins), [team](https://github.com/fabric8-services/fabric8-tenant-team)) to the directory `environment/templates/` inside of this `fabric8-tenant` project. The files (templates) located in the directory and their names correspond to namespaces they are used for. All objects (contained in the templates) have been moved from the original ones (downloaded from maven central) just with small cosmetic changes - the result and the number of the objects are same.
The cosmetic changes are:
* some of the annotations that were used by FMP are removed
* the version of the templates corresponds to the latest commit of  `fabric8-tenant` project
* labels were slightly changed to not look like a java project
* the content of `fabric8-kubernetes-user-sa.yml` was removed as it wasn't used anywhere
* there is only one template for both `run` and `stage` namespaces: `fabric8-tenant-deploy.yml` - the name of the deploy environment is set by `DEPLOY_TYPE`
* objects of the templates have been rearranged to reflect the namespaces they are used for and files they are located in

Originally, the in-production testing of the changed templates was done via PRs and by setting a release (Nexus) version in https://openshift.io/_profile/_tenant page. The in-production testing will be still possible just with a small change - it won't be downloaded from Nexus, but directly from the GH PR.  That will be possible by taking `sha` of the latest commit in the PR and setting it on the same page as it is done currently. It will be also possible to change the repo and directory the files are downloaded from so you will not have to create a PR to test a template in production.

For better verification, I captured the requests that are sent to openshift api when the namespaces are being created. I printed out the URL and the body of each request.
* this is content for the original version (from master): https://gist.github.com/MatousJobanek/12d062712410349d1e980369a71f6181
* this is content for the code from this PR: https://gist.github.com/MatousJobanek/3beefc66c8e98b147e68dd9f04188662
you can compare them and see that the only differences are labels, annotations, and version.

TODO:
- [x] change https://openshift.io/_profile/_tenant page to reflect the new parameters - PR is here https://github.com/fabric8-ui/fabric8-ui/pull/3344
[...]

PR #648, opened 2018-10-03
Use sha of latest commit that changes a template file as its version

MatousJobanek:
This PR is a followup of a PR https://github.com/fabric8-services/fabric8-tenant/pull/634. It changes the template versioning system to a form we agreed on in a mailing thread - that as a template version a `sha` of the latest commit that changes the particular file is used. This is set by `LDFLAGS`. The correct creation of the `LDFLAGS` is verified by a test that is executed as part of `test-unit` target:

### DIFF
--- a/environment/service_test.go
+++ b/environment/service_test.go
@@ -7,6 +7,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/h2non/gock.v1"
+	"regexp"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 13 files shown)


environment/service_test.go
```diff
@@ -0,0 +1,196 @@
...
+	// given
+	service := environment.NewService("", "", "")
+	setTemplateVersions()
+	vars := map[string]string{
+		"USER_NAME": "dev",
+	}
+
+	for _, envType := range environment.DefaultEnvTypes {
+		// when
+		env, err := service.GetEnvData(context.Background(), envType)
```
golangcibot: ineffectual assignment to `err`

MatousJobanek: fixed and rebased to one commit 2cb58ab

MatousJobanek:
[test]

environment/service_test.go
```diff
@@ -39,10 +41,16 @@ objects:
 func TestGetAllTemplatesForAllTypes(t *testing.T) {
```
alexeykazakov: Let's add a check in SetupSuite (so, it's run before any test) that version/commit vars are set (!= "") before you set them to test values in your tests.

alexeykazakov: Never mind. We won't have these vars set in tests. So, can't test it here.

alexeykazakov:
Approving it but we still need to solve the problem with independent versions of Che and Che-quota & Jenkins and Jenkins-quota templates. Can be done in a following PR though.

alexeykazakov:
> I also see the following when running make test-templates-flags locally:

This PR fixes it: https://github.com/MatousJobanek/fabric8-tenant/pull/2

PR #653, opened 2018-10-05
update README to reflect changes in template management & add Dockerfile.builder.local

MatousJobanek:
* updates README to reflect changes in template management describing versioning system and in-production testing
* adds `Dockerfile.builder.local` and related target `docker-start-local` in makefile for running integration tests locally.


Dockerfile.builder.local
```diff
@@ -0,0 +1,26 @@
...
+      which \
+    && yum clean all
+
+# Get dep for Go package management
+RUN mkdir -p /tmp/go/bin
+ENV GOPATH /tmp/go
+RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh && mv /tmp/go/bin/dep /usr/bin
+
+RUN chmod a+rwx ${HOME}
+RUN chmod -R a+rwx ${GOPATH}
```
alexeykazakov: Why not just adding it to Dockerfile.builder? Why introducing Dockerkfile.builder.local?
Also wondering why `chmod a+rwx ${HOME}` and `chmod -R a+rwx /tmp` are required here. `RUN chmod -R a+rwx ${GOPATH}` is enough in our other repos like wit, auth, cluster, etc.

MatousJobanek: > Why not just adding it to Dockerfile.builder? Why introducing Dockerkfile.builder.local?

because you asked me to do that some time ago :-) - but that was just caused by misunderstanding.

> Also wondering why chmod a+rwx ${HOME} and chmod -R a+rwx /tmp are required here. RUN chmod -R a+rwx ${GOPATH} is enough in our other repos like wit, auth, cluster, etc.

yeah, you are right, I have no idea why it didn't work without setting the permissions also for the other directories...

I've cleaned up the mess here: 9299872

PR #659, opened 2018-10-12
WIP: Create contract tests for interaction with Auth service.

pmacik:


.make/test.mk
```diff
@@ -117,6 +117,14 @@ ALL_PKGS_EXCLUDE_PATTERN = 'vendor\|app\|tool\/cli\|design\|client\|test'
 GOANALYSIS_PKGS_EXCLUDE_PATTERN="vendor|app|client|tool/cli"
 GOANALYSIS_DIRS=$(shell go list -f {{.Dir}} ./... | grep -v -E $(GOANALYSIS_PKGS_EXCLUDE_PATTERN))
 
+# Folder with contract tests
+CONTRACT_TESTS=$(CUR_DIR)/test/contracts
+
+# Configuration of contract tests
+PACT_VERSION ?= 1.0.0
+PACT_BROKER_URL ?= http://pact-broker-pact-broker.193b.starter-ca-central-1.openshiftapps.com
```
alexeykazakov: We can't relay on OSO-starter to run the broker. Need a better place for that.

bartoszmajsak: Maybe the same home as sentry and ike bots? /cc @kbsingh

bartoszmajsak:
Who is a provider and who is a consumer here?

bartoszmajsak:
Few other points

- publishing part ideally should be part of the test code - so we can nicely wrap it as part of test execution rather than a shell script run afterwards
- I would think of contract tests actually touching some real component talking with the underlying service - thinking about it as a "shareable mock" - you mock auth service but excercise some actual logic in the component which communicates with it

.make/test.mk
pmacik: @alexeykazakov @bartoszmajsak Thanks for the suggestion, created appropriate SD ticket  for that (https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2383)

pmacik:
> Who is a provider and who is a consumer here? Or in other words - where can I see auth tests?

@bartoszmajsak The provider is the Auth service (`PACT_PROVIDER_BASE_URL`) while the consumer is the tenant service itself.

It's still a WIP so both contract sides are in the same place, but the auth tests (the provider tests) are the part verifying the pact file taken from broker against a living auth service instance (=`test-contract-auth-verify-broker`, I plan to move that to the auth service repo and include it in their tests).

PR #662, opened 2018-10-16
chore: improve sentry errors

MatousJobanek:
There are two small improvements:

* in case of `NewUnauthorizedError` there was reported only the text `Could not resolve user token` without any additional information. I've attached the error that caused the failure to the message.

* in case of other errors than the internal ones, there was reported `error.stringError` instance. I've replaced it by wrapping the created message into the original error


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-tenant/blob/58c91d771d849a9d0c6bfbd8e5d32728693109fa/.ike-prow/test-keeper.yaml).

PR #665, opened 2018-10-17
Issue #4020 Allow isDirectory() operation

kishansagathiya:
Allow java.io.File.isDirectory() so that we would be able to check in
the pipeline if we have create nodes directory

https://github.com/openshiftio/openshift.io/issues/4020


kishansagathiya:
@chmouel I have removed `mkdirs` as users can modify pipeline library and create random repos in jenkins using that. However `isDirectory` gives only read access. Read excess we have already given by `exists`, this is similar to that. 

I can't think of scenarios where someone can misuse isDirectory in order to create a security threat.

Going through commits, sadly none of the commits provide any info on why certain operations were not allowed or why they were permitted later.

chmouel:
@kishansagathiya thanks for doing this investigation, it wasn't as policy previously (but we will be doing a review), can you as well please update with your findings on MM,

PR #666, opened 2018-10-17
Re add ConfigChange trigger on Jenkins DC

aslakknutsen:


chmouel:
@aslakknutsen so we do need it? was that because tenant update is not working without it ?

aslakknutsen:
Yea, it didn't work as thought anyway. So essentially just revert for now

aslakknutsen:
as in it's causing more problems then good :)

PR #668, opened 2018-10-31
chore: move whoami logic to cluster

MatousJobanek:
This PR moves the "whoami" logic to `cluster` package. There are two reasons:

* the logic is used only there
* avoid cyclic package dependency in future


MatousJobanek:
[test]

MatousJobanek:
[test]

MatousJobanek:
[test]

PR #673, opened 2018-11-01
chore: refactor core functionality to be more extensible

MatousJobanek:
This PR refactors the core logic of the tenant service that is mainly placed in `openshift` and `controller` packages. As part of it it also  changes the way the update and delete is done. 
For a better understanding of the changes (mainly in the `openshift` package), I would recommend to fetch this PR and follow the code starting from `main.go` or `controller` package, not checking the changes in GH web UI as most of the code is totally different.

#### Why this PR

In the original code there were code [duplications](https://github.com/fabric8-services/fabric8-tenant/blob/master/controller/tenant.go#L271); several complicated [if-else](https://github.com/fabric8-services/fabric8-tenant/blob/master/controller/tenant.go#L271) statements to handle [exceptions](https://github.com/fabric8-services/fabric8-tenant/blob/master/openshift/init_tenant.go#L32) in the behavior; inconsistent way in creating/updating/deleting objects in OS, etc. It was really hard to extend the logic by new features and make it ready for OSIO on OSD

#### Design

The new design I'm proposing in this PR encapsulates the logic in objects, which means that every environment type or action performs what is needed without any complicated if-else statements. The approach could be understood as: 
"We need to perform an action(eg. POST) on an environment type (eg. Che) for a particular tenant. To do this, we need to apply a list of template objects (specific for the type) using OpenShift endpoints. Every object (eg. projectrequest, rolebinding, ...) has a different endpoint and for every HTTP method may have different strategy for handling the request or possible errors."
The new and most important objects proposed in the PR are:

* `openshift.NamespaceAction` - Represents the action that should be applied on the namespaces for the particular tenant - [post|update|delete]. It is mainly responsible for operation on DB and provides additional information specific to the action that is needed by other objects
* `cluster.ForType` - it is not a object but a function that returns cluster for environment type. This is kind of preparation for the multicluster environment - in other words, it represents cluster mapping per env-type.
* `openshift.EnvironmentTypeService` - Represents service operating with information related to environment types(template, objects, cluster,...). It is responsible for getting and filtering objects to be applied, provides information (eg. needed token) specific for the particular type and performs after-apply-callback (needed by user namespace)
* `openshift.ObjectEndpoints` - is list of MethodDefinitions for a particular object endpoint (eg. `/oapi/v1/projectrequests`). In other words, is saying which methods (Post/Delete/Get/Patch) are allowed to be performed for the endpoint
* `openshift.MethodDefinition` - represents defined actions (beforeDoCallbacks, afterDoCallbacks,requestCreator) to be executed when the method is performed for an endpoint.
* `openshift.Service` - knowing which action is requested it starts for every environment type a new goroutine. The goroutine gets template objects to be applied and the target cluster and then sends a request for every object to the OS cluster.

#### The flow of every action

The flow of every action is described in comments of every action - the resulting flow should reflect the original implementation

#### Benefits of the changes

* since the specific logic is encapsulated inside of the objects, then it is much easier eg. to introduce any new environment (test, custom) by just adding a new implementation of the type
* using the HTTP requests for all actions (update and delete) instead of oc commands gives us a possibility to easily handle errors (which is badly needed for PVC that are in terminating state for a long time) and should be also faster 
* since the method `ApplyAll(nsTypes ...environment.Type)` takes any number of env-types it prepares the logic for lazy initialization of single namespace. For the lazy-init also the function `filterMissingAndExisting(namespaces []*tenant.Namespace)` will be very useful
* the `cluster.ForType` mapping prepares the code for multi-cluster environment
* it will be much easier to introduce support for contributors in OSD spaces (creation of namespaces in spaces on OSD cluster)

Fixes https://github.com/fabric8-services/fabric8-tenant/issues/678


tenant/repository_test.go
```diff
@@ -155,7 +157,12 @@ func (s *TenantServiceTestSuite) TestDelete() {
 		tenant1 := fxt.Tenants[0]
 		tenant2 := fxt.Tenants[1]
 		// when
-		svc.DeleteAll(tenant1.ID)
+		namespaces, err := svc.GetNamespaces(tenant1.ID)
+		require.NoError(t, err)
+		for _, ns := range namespaces {
+			svc.NewTenantRepository(tenant1.ID).DeleteNamespace(ns)
```
golangcibot: Error return value of `(github.com/fabric8-services/fabric8-tenant/tenant.Repository).DeleteNamespace` is not checked

tenant/repository_test.go
```diff
@@ -155,7 +157,12 @@ func (s *TenantServiceTestSuite) TestDelete() {
...
+		}
+		svc.DeleteTenant(tenant1.ID)
```
golangcibot: Error return value of `svc.DeleteTenant` is not checked

utils/url.go
```diff
@@ -24,3 +25,22 @@ func DownloadFile(url string) ([]byte, error) {
...
 }
+
+func ReadBody(response *http.Response) (func() []byte, func()) {
+	if response == nil {
+		return func() []byte {
+			return []byte{}
+		}, func() {}
+	}
+	toDefer := func() {
+		ioutil.ReadAll(response.Body)
```
golangcibot: Error return value of `ioutil.ReadAll` is not checked

utils/url.go
```diff
@@ -24,3 +25,22 @@ func DownloadFile(url string) ([]byte, error) {
...
+		response.Body.Close()
+	}
+
+	readBody := func() []byte {
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(response.Body)
```
golangcibot: Error return value of `buf.ReadFrom` is not checked

openshift/action.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (d *Delete) filter() FilterFunc {
+	if d.removeFromCluster {
+		return isOfKind(environment.ValKindProjectRequest)
+	}
+	return isOfKind(environment.ValKindPersistenceVolumeClaim, environment.ValKindConfigMap)
+}
+
+func (d *Delete) sort(toSort environment.ByKindSorter) {
+	sort.Reverse(toSort)
```
golangcibot: result of sort.Reverse call not used

openshift/endpoints.go
```diff
@@ -0,0 +1,204 @@
...
+			endpoint(`/api/v1/namespaces/{{ index . "metadata" "namespace"}}/resourcequotas/{{ index . "metadata" "name"}}`, PATCH(), GET(), DELETE())),
+
+		environment.ValKindLimitRange: endpoints(
+			endpoint(`/api/v1/namespaces/{{ index . "metadata" "namespace"}}/limitranges`, POST(WhenConflictThenDeleteAndRedo)),
+			endpoint(`/api/v1/namespaces/{{ index . "metadata" "namespace"}}/limitranges/{{ index . "metadata" "name"}}`, PATCH(), GET(), DELETE())),
+
+		environment.ValKindJob: endpoints(
+			endpoint(`/apis/batch/v1/namespaces/{{ index . "metadata" "namespace"}}/jobs`, POST(WhenConflictThenDeleteAndRedo)),
+			endpoint(`/apis/batch/v1/namespaces/{{ index . "metadata" "namespace"}}/jobs/{{ index . "metadata" "name"}}`, PATCH(), GET(), DELETE())),
+	}
```
nurali-techie: @MatousJobanek one of the basic/major design feedback I have on `endpoint` object.

I think the _old arrangement_ was simple [code_link](https://github.com/fabric8-services/fabric8-tenant/blob/master/openshift/apply.go#L28) **where endpoints are grouped for given action** (POST/PATCH)
rather, this _new arrangement_ **where endpoints are tagged with action it supports**.

In old_code, I really don't like `recursive` call to apply() [code_link](https://github.com/fabric8-services/fabric8-tenant/blob/master/openshift/apply.go#L233) which was making it hard/complex to understand.

In new_code, I like the way code is reused :-)

old one goes like this (in the code has definition of sequence_flow) .. if you want to `Setup()` tenant, `POST` is the action and for POST these are list_of_endpoints.  Call them one by one and response of each endpoint has to be further examined to take appropriate steps .. this helps in debugging the issue. for ex, we got issue which say when I provision_tenant, I get so and so error.  To debug just open code and see what are the endpoints called for `POST` and follow the execution of each endpoint.

So overall, see if you can fix endpoint_arrangement while still keep rest of design which provides code_reusability.  If it's a lot_work than you can take one more feedback from team and see if other feel the this problem or not.

MatousJobanek: @nurali-techie thank you very much for your feedback and sorry for may delayed response.

> I think the old arrangement was simple code_link where endpoints are grouped for given action (POST/PATCH) rather, this new arrangement where endpoints are tagged with action it supports.

I see your point - I think that it is just a question of the point of view. The old code was grouped by the action/method, but was sometimes hard to find all endpoints for one object kind. And also to check that there are all endpoints set for all actions for a particular object kind. Now all information are grouped together and visible in one place.

> So overall, see if you can fix endpoint_arrangement while still keep rest of design which provides code_reusability. If it's a lot_work then you can take one more feedback from team and see if other feel the this problem or not.

I could do that, but the only thing that would be changed is the endpoints arrangement. The internal logic would stay the same (because I rely on it in other places) so I'm not sure if that would anyhow help with debugging. I'm not sure how much time that would need, maybe a few hours, maybe one day. Let's wait for the other guys what they think about it.

fabric8cd:
This PR changes one or more of the template files. These changes are now available for testing: Launch in [OpenShift.io](https://openshift.io/_profile/_tenant?templatesRepoBlob=7b05db8406589bf124bef454e04abaae38e45e38) and click the update tenant button

fabric8cd:
This PR changes one or more of the template files. These changes are now available for testing: Launch in [OpenShift.io](https://openshift.io/_profile/_tenant?templatesRepoBlob=79e87daa39c97a4afb7d544be0e91037597737f4) and click the update tenant button

environment/service.go
```diff
@@ -83,30 +83,55 @@ type Service struct {
...
-		templatesRepoBlob: templatesRepoBlob,
-		templatesRepoDir:  templatesRepoDir,
+func NewService() *Service {
+	return &Service{}
+}
+
+func NewServiceForUserData(user *authclient.UserDataAttributes) *Service {
+	service := NewService()
+	if user != nil {
+		if user.FeatureLevel != nil && *user.FeatureLevel != "internal" {
```
xcoulon: we should have a constant for this `internal` feature level ;)

openshift/action.go
```diff
@@ -0,0 +1,350 @@
...
+				"self-healing-strategy": "recreate-with-new-nsBaseName",
+			}, "the creation failed, starting self-healing logic")
+			openShiftUsername := openShiftService.service.context.openShiftUsername
+			tnnt, err := c.tenantRepo.GetTenant()
+			errMsgSuffix := fmt.Sprintf("while doing self-healing operations triggered by error: [%s]", originalError)
+			if err != nil {
+				return errors.Wrapf(err, "unable to get tenant %s", errMsgSuffix)
+			}
+			namespaces, err := c.tenantRepo.GetNamespaces()
+			fmt.Println(namespaces[0].ID)
```
xcoulon: please remove this statement

openshift/action.go
```diff
@@ -0,0 +1,350 @@
...
+			err = openShiftService.WithPostMethod(false).ApplyAll(environment.DefaultEnvTypes)
+			if err != nil {
+				return errors.Wrapf(err, "unable to create new namespaces %s", errMsgSuffix)
+			}
+			return nil
+		}
+	}
+}
+
+func NewCreate(tenantRepo tenant.Repository, allowSelfHealing bool) *Create {
```
xcoulon: Maybe rename `Create` to `CreateAction` ? `Create` alone looks a bit strange IMO

tenant/repository.go
```diff
@@ -21,8 +21,9 @@ type Service interface {
 	// SaveTenant will update on dupliate 'insert'
```
xcoulon: `duplicate`

tenant/repository.go
```diff
@@ -116,6 +117,36 @@ func (s DBService) SaveNamespace(namespace *Namespace) error {
...
+	if namespace.ID == uuid.Nil {
+		namespace.ID = uuid.NewV4()
+	}
+	length := len(namespace.Name)
+	lockid := length + int([]rune(namespace.Name)[length-1])
+	created := false
+	err := dbsupport.Transaction(s.db, dbsupport.Lock(lockid, 10, func(tx *gorm.DB) error {
+		var duplicate []*Namespace
+		err := tx.Table(namespaceTableName).
+			Where("name = ? AND master_url = ? AND tenant_id = ?", namespace.Name, namespace.MasterURL, namespace.TenantID).
```
xcoulon: have you checked that the table has an index on those 3 columns?

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
 	return s.db.Unscoped().Delete(&Tenant{ID: tenantID}).Error
 }
 
-type NilService struct {
+func (s DBService) NewTenantRepository(tenantID uuid.UUID) Repository {
+	return DBTenantRepository{service: s, tenantID: tenantID}
```
xcoulon: why does the repository need to handle a reference to the service?

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
...
+}
+
+func (n DBTenantRepository) GetTenant() (*Tenant, error) {
+	return n.service.GetTenant(n.tenantID)
 }
 
-func (s NilService) Exists(tenantID uuid.UUID) bool {
-	return false
+func (n DBTenantRepository) GetNamespaces() ([]*Namespace, error) {
+	return n.service.GetNamespaces(n.tenantID)
```
xcoulon: so this function should not belong to the repository?

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
...
+func (n DBTenantRepository) SaveNamespace(namespace *Namespace) error {
+	namespace.TenantID = n.tenantID
+	return n.service.SaveNamespace(namespace)
 }
 
-func (s NilService) GetNamespaces(tenantID uuid.UUID) ([]*Namespace, error) {
-	return nil, nil
+func (n DBTenantRepository) CreateNamespace(namespace *Namespace) (*Namespace, error) {
+	namespace.TenantID = n.tenantID
+	return n.service.CreateNamespace(namespace)
```
xcoulon: same here: no need to have this method here?

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
...
 }
 
-func (s NilService) SaveTenant(tenant *Tenant) error {
-	return nil
+func (n DBTenantRepository) DeleteNamespace(namespace *Namespace) error {
+	return n.service.db.Unscoped().Delete(namespace).Error
```
xcoulon: same as above

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
...
 }
 
-func (s NilService) SaveNamespace(namespace *Namespace) error {
-	return nil
+func (n DBTenantRepository) DeleteNamespaces() error {
+	return n.service.DeleteNamespaces(n.tenantID)
```
xcoulon: same as above

tenant/repository.go
```diff
@@ -162,35 +193,69 @@ func (s DBService) DeleteTenant(tenantID uuid.UUID) error {
...
 }
 
-func (s NilService) LookupTenantByClusterAndNamespace(masterURL, namespace string) (*Tenant, error) {
-	return nil, nil
+func (n DBTenantRepository) DeleteTenant() error {
+	return n.service.DeleteTenant(n.tenantID)
```
xcoulon: same as above

xcoulon:
Good job overall, and congratulations for raising the test coverage by 13.03%! 🙌💪

cluster/service.go
```diff
@@ -87,6 +90,23 @@ func (s *clusterService) Start() error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return func(envType environment.Type) Cluster {
+		return cluster
+	}, nil
+}
+
+func ForTypeMapping(mapping map[environment.Type]Cluster) ForType {
```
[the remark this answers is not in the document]
MatousJobanek: added 18eafc9

cluster/service.go
```diff
@@ -87,6 +90,23 @@ func (s *clusterService) Start() error {
 	return nil
 }
 
+func (s *clusterService) GetUserClusterForType(ctx context.Context, user *auth.User) (ForType, error) {
```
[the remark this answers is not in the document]
MatousJobanek: added 18eafc9

cluster/service.go
```diff
@@ -28,11 +29,13 @@ type Cluster struct {
 }
 
 type GetCluster func(ctx context.Context, target string) (Cluster, error)
+type ForType func(envType environment.Type) Cluster
```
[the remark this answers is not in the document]
MatousJobanek: added 18eafc9

controller/convert_whitebox_test.go
```diff
@@ -0,0 +1,114 @@
...
+			MasterURL: "http://test2.org",
+			Name:      "test-jenkins",
+			Type:      environment.TypeJenkins,
+			Version:   "1.0",
+			State:     "created",
+		},
+	}
+}
+
+func Test_convertTenant(t *testing.T) {
```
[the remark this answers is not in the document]
MatousJobanek: blindly copied from the original location - fixed b28b408

controller/tenant.go
```diff
@@ -2,314 +2,379 @@ package controller
...
-	tenantService tenant.Service,
+	tenantRepository tenant.Service,
 	clusterService cluster.Service,
 	authClientService auth.Service,
 	config *configuration.Data) *TenantController {
 
 	return &TenantController{
-		Controller:        service.NewController("TenantController"),
-		tenantService:     tenantService,
+		Controller:        service.NewController("TenantsController"),
```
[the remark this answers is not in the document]
MatousJobanek: copy-paste driven development :-). fixed 54c8447
good catch :+1:

controller/tenant.go
```diff
@@ -2,314 +2,379 @@ package controller
...
-	}
+	TenantUpdater{Config: c.config, ClusterService: c.clusterService, TenantRepository: c.tenantRepository}.
+		Update(ctx, dbTenant, user, environment.DefaultEnvTypes, true)
 
-	cluster, err := c.clusterService.GetCluster(ctx, *user.UserData.Cluster)
+	ctx.ResponseData.Header().Set("Location", rest.AbsoluteURL(ctx.RequestData.Request, app.TenantHref()))
+	return ctx.Accepted()
+}
+
+type TenantUpdater struct {
```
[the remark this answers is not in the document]
MatousJobanek: What do you by that? I don't follow... What would be the benefit?

controller/tenant.go
```diff
@@ -2,314 +2,379 @@ package controller
...
-		_, err = openshift.RawInitTenant(ctx, openshiftConfig, t, user.OpenShiftUserToken, c.tenantService, true)
+	// creates openshift services
+	openShiftService := c.newOpenShiftService(ctx, user, dbTenant.NsBaseName, clusterMapping)
 
-		if err != nil {
-			sentry.LogError(ctx, map[string]interface{}{
-				"os_user": user.OpenShiftUsername,
-			}, err, "unable initialize tenant")
+	// perform delete method on the list of existing namespaces
+	err = openShiftService.WithDeleteMethod(namespaces, removeFromCluster, !removeFromCluster, true).ApplyAll(environment.DefaultEnvTypes)
```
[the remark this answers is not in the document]
MatousJobanek: I know, I don't like it either. 
* first param are existing namespaces
* second is if it should be removed from the cluster - in other words, if we should just clean the content or remove it completely
* third is if we should keep tenant entity in DB or remove it - when removal is done
* fourth is if the self-healing strategy is enabled 

how about this:
if we should only clean the content (this means that also tenant entity has to stay in DB):
```go
if removeFromCluster {
	err = openShiftService.WithDeleteMethod(namespaces, true).RemoveFromCluster().ApplyAll(environment.DefaultEnvTypes)
} else {
	err = openShiftService.WithDeleteMethod(namespaces, true).ApplyAll(environment.DefaultEnvTypes)
}
```
where first param of `WithDeleteMethod` are existing namespaces and second information if self-healing is activated. When `RemoveFromCluster()` is called then, by default, the entity from DB would be removed. When it is required to keep the entity, then we should call:
```go
err = openShiftService.WithDeleteMethod(namespaces, true).RemoveFromCluster().ButKeepTenantEntity().ApplyAll(environment.DefaultEnvTypes)
``` 
does it make sense?

controller/update_test.go
```diff
@@ -112,32 +117,45 @@ func (s *UpdateControllerTestSuite) TestStartUpdateOk() {
 
 		// then
 		err := test.WaitWithTimeout(10 * time.Second).Until(func() error {
-			if int(*updateExecutor.NumberOfCalls) != 60 {
+			if int(*updateExecutor.NumberOfCalls) != 12 {
 				return fmt.Errorf("expeced number of calls 60 wasn't fullfiled - actual: %d", int(*updateExecutor.NumberOfCalls))
```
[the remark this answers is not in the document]
MatousJobanek: Because the original logic was calling the executor for every namespace separately (which means five times per one tenant) - this new logic calls executor only once for one tenant and gives a list of env types to be updated - the update of the actual namespaces is managed by the executor

environment/service.go
MatousJobanek: done c60a726

openshift/action.go
MatousJobanek: removed 822ba3b

openshift/action.go
MatousJobanek: renamed 822ba3b (the other action structs as well)

tenant/repository.go
MatousJobanek: Because the repository is just a wrapper (that keeps the `tenantID`) and calls the service. 
To be honest, in the first stages of my work I also totally changed this representation of the access to the DB. But as part of my attempt to keep the PR as small as possible, I didn't introduce this change here. I just created the wrapper so I don't have to pass the `tenantID` as a parameter everywhere, so I ended up with the approach of having the service as the source of all logic and using the repo just as the wrapper. 
However, during the time the PR went bigger and bigger and I should have changed this repo/service solution as well... I'll try to move the relevant logic from the service to the repository and I'll change also the places where it is used - or, at least, I'll try my best ;-)

tenant/repository.go
MatousJobanek: fixed as part of 1c4a496

tenant/repository.go
MatousJobanek: ok, I tried my best and this is the result: 1c4a496 I hope that it looks better now

tenant/repository.go
MatousJobanek: should be solved as part of 1c4a496

tenant/repository.go
MatousJobanek: should be solved as part of 1c4a496

tenant/repository.go
MatousJobanek: should be solved as part of 1c4a496

tenant/repository.go
MatousJobanek: should be solved as part of 1c4a496

tenant/repository.go
MatousJobanek: should be solved as part of 1c4a496

MatousJobanek:
> Also, regarding the openshift package, I wonder if we could replace some parts with https://github.com/openshift/client-go ?

That was one of my first ideas - to use openshift/k8s clients for handling the communication with OS. I tried to use it in my code, but I came across a few issues - there are two most important ones:
* As you probably noticed, the tenant code handles the response codes and based on the values it reacts in various ways. That wasn't possible with the clients - the API wraps the calls and hides the response codes, so it is hard to figure out what failed, why it failed (if anything failed) etc... something similar as we have with the `oc` client now. There is maybe some way of getting the response codes from the clients, but I wasn't able to figure it out. 
* At the time when I was playing with it (September 2018) there was missing client support for at least one object kind (I cannot recall which one - maybe `RoleBindingsRestriction` or `ResourceQuota`)

So instead of mixing two ways of doing that (because of missing client support for some object kinds) and trying to workaround/hack the missing response code information, I decided to go with one consistent way that is proposed in this PR.

Just to be clear - I'm definitely not against using it and if we figure out that the go-clients offer us everything we need, we can replace it. I believe that it shouldn't be so hard with my code as everything is hidden behind the endpoints/method abstraction.

> Good job overall, and congratulations for raising the test coverage by 13.03%! raised_handsmuscle

Thanks. Actually, the increase was originally higher - at one time even around 16-17%, but went lower as I introduced a few tests separately. :-)

xcoulon:
> > Also, regarding the openshift package, I wonder if we could replace some parts with https://github.com/openshift/client-go ?
> 
> That was one of my first ideas - to use openshift/k8s clients for handling the communication with OS. I tried to use it in my code, but I came across a few issues - there are two most important ones:
> 
> * As you probably noticed, the tenant code handles the response codes and based on the values it reacts in various ways. That wasn't possible with the clients - the API wraps the calls and hides the response codes, so it is hard to figure out what failed, why it failed (if anything failed) etc... something similar as we have with the `oc` client now. There is maybe some way of getting the response codes from the clients, but I wasn't able to figure it out.
> * At the time when I was playing with it (September 2018) there was missing client support for at least one object kind (I cannot recall which one - maybe `RoleBindingsRestriction` or `ResourceQuota`)
> 
> So instead of mixing two ways of doing that (because of missing client support for some object kinds) and trying to workaround/hack the missing response code information, I decided to go with one consistent way that is proposed in this PR.
> 

ok, I didn't know there were some missing parts. Maybe we could open one or more issues on the upstream repo if we really want to use this lib (that is, if other problems that you encountered can be lifted as well)

> Just to be clear - I'm definitely not against using it and if we figure out that the go-clients offer us everything we need, we can replace it. I believe that it shouldn't be so hard with my code as everything is hidden behind the endpoints/method abstraction.
> 
Yes, that's my thinking as well, and my point is we could have a simpler wrapper around the library. But again, it depends if the client library has all the features that we need

> > Good job overall, and congratulations for raising the test coverage by 13.03%! raised_handsmuscle
> 
> Thanks. Actually, the increase was originally higher - at one time even around 16-17%, but went lower as I introduced a few tests separately. :-)
Yeah, well, this is still a significant increase ;)

controller/tenant.go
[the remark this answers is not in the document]
xcoulon: what about going a step further with the self-healing strategy support as well, or something slightly different (not using a builder but rather with "options" via a variadic func):
```
err = openshiftService.ApplyAll(namespaces, environment.DefaultEnvTypes) 
```
in which case, some default settings would be applied (eg: just clean the tenant, etc.)
then, with options:
```
err = openshiftService.ApplyAll(namespaces, environment.DefaultEnvTypes, DeleteTenant, EnableSelfHealing) 
```
see what I mean? Does it make sense?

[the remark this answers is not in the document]
MatousJobanek: I see your point. However, this cannot be used for the setup/creation - there are no namespaces needed, or even expected.  
So how about a combination of both:
```go
openShiftService.WithDeleteMethod(namespaces, EnableSelfHealing).ApplyAll(environment.DefaultEnvTypes)

openShiftService.WithDeleteMethod(namespaces, EnableSelfHealing, RemoveFromCluster).ApplyAll(environment.DefaultEnvTypes)

openShiftService.WithDeleteMethod(namespaces, RemoveFromCluster, KeepTenantEntity).ApplyAll(environment.DefaultEnvTypes)
```

tenant/repository.go
xcoulon: thanks @MatousJobanek this is much better indeed! 🙌

tenant/repository.go
xcoulon: ok, thanks

cluster/service.go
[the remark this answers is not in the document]
xcoulon: thanks!

cluster/service.go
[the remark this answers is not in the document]
xcoulon: Thanks!

cluster/service.go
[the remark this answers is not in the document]
xcoulon: Thanks!

controller/convert_whitebox_test.go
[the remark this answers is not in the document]
xcoulon: ok ;)

controller/tenant.go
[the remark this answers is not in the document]
xcoulon: ok ;)

tenant/repository.go
xcoulon: Thanks!

tenant/repository.go
xcoulon: Thanks!

tenant/repository.go
xcoulon: Thanks!

tenant/repository.go
xcoulon: Thanks!

controller/tenant.go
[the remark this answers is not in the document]
MatousJobanek: what `TenantService` you are referring to?

controller/tenant.go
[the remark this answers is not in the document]
MatousJobanek: Does it look better? 732e5bc

controller/tenant.go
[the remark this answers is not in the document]
xcoulon: oh, I thought there was already a `TenantService`. In anyways, this would probably be a good place for this code, I believe. But that could be done in a separate PR if you prefer

controller/tenant.go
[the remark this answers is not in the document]
xcoulon: yes, it's much more readable now! Thanks for doing this change!

fabric8cd:
This PR changes one or more of the template files. These changes are now available for testing: Launch in [OpenShift.io](https://openshift.io/_profile/_tenant?templatesRepoBlob=25c7303efb4f5eaef9eb48e92dd9ab52c186f2a7) and click the update tenant button

Issue #674, opened 2018-11-01
Conflict when two users with same name are registered


MatousJobanek:
consider users:
johny@first.com
johny@second.com
When the first johny is registered, then the base-name for namespaces is `johny` and all namespaces will be created using this base-name - which means `johny`, `johny-che`, `johny-jenkins`, `johny-run` and `johny-stage`

However when the second johny is registered, then same base-name is used and tries to create namespaces that already exists which fails due to the rolebindingrestriction.
see https://errortracking.prod-preview.openshift.io/openshift_io/fabric8-tenant/issues/9033/events/385234/

PR #675, opened 2018-11-01
fix(#674): introduce nsUsername used as base-name for namespace names

MatousJobanek:
This PR introduces `nsUsername` in `tenant` table. It is used as a base-name (eg. `johny`) for namespace names. When a new tenant is being created, then it checks if there is some tenant or namespace that the same base-name already uses. If there is some, then it appends a number to the name starting with number 2 (eg. `johny2`) and checks for any conflict again.

### DIFF
--- a/controller/tenant.go
+++ b/controller/tenant.go
@@ -83,12 +83,15 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
 	}
 
+	username := tenant.ConstructNsUsername(c.tenantService, env.RetrieveUserName(user.OpenShiftUsername))
+
 	// create openshift config
 	openshiftConfig := openshift.NewConfig(c.config, user.UserData, cluster.User, cluster.Token, cluster.APIURL)
 	tenant := &tenant.Tenant{
 		ID:         ttoken.Subject(),
 		Email:      ttoken.Email(),
 		OSUsername: user.OpenShiftUsername,
+		NsUsername: username,
 	}
 	err = c.tenantService.CreateTenant(tenant)
 	if err != nil {
@@ -104,8 +107,9 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
 		err = openshift.RawInitTenant(
 			ctx,
 			openshiftConfig,
-			InitTenant(ctx, openshiftConfig.MasterURL, c.tenantService, t, user.OpenShiftUsername),
+			InitTenant(ctx, openshiftConfig.MasterURL, c.tenantService, t),
 			user.OpenShiftUsername,
+			username,
 			user.OpenShiftUserToken)
 
 		if err != nil {
@@ -156,7 +160,9 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
 
 	// update tenant config
 	tenant.OSUsername = user.OpenShiftUsername
-
+	if tenant.NsUsername == "" {
+		tenant.NsUsername = env.RetrieveUserName(user.OpenShiftUsername)
+	}
 	if err = c.tenantService.SaveTenant(tenant); err != nil {
 		log.Error(ctx, map[string]interface{}{
 			"err": err,
@@ -170,8 +176,9 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
 		err = openshift.RawUpdateTenant(
 			ctx,
 			openshiftConfig,
-			InitTenant(ctx, openshiftConfig.MasterURL, c.tenantService, t, user.OpenShiftUsername),
-			user.OpenShiftUsername)
+			InitTenant(ctx, openshiftConfig.MasterURL, c.tenantService, t),
+			user.OpenShiftUsername,
+			tenant.NsUsername)
 
 		if err != nil {
 			sentry.LogError(ctx, map[string]interface{}{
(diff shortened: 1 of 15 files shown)


migration/sql-files/006-add-ns-username-column-to-tenant.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE tenants
+ADD COLUMN ns_username TEXT;
```
aslakknutsen: Index?

aslakknutsen: Maybe add a migration to add the basename of the existing users. Using the namespace.name where type=user should work..

MatousJobanek: index added: https://github.com/fabric8-services/fabric8-tenant/pull/675/commits/a1914bc2b6b7aca004f1a9d8eb91a305ff0cb5ed
is it fine that it is in the same file? or did you expect another migration file?

MatousJobanek: Actually, it's good idea - with this we could easily find those tenants that are not properly provisioned - have no `user` namespace

aslakknutsen: Yeah, easy to find anyway, but probably about 20 at this point I think

tenant/repository.go
```diff
@@ -39,6 +42,15 @@ func (s DBService) Exists(tenantID uuid.UUID) bool {
 	return true
 }
 
+func (s DBService) ExistsWithNsUsername(nsUsername string) bool {
+	var t Tenant
+	err := s.db.Table(t.TableName()).Where("ns_username = ?", nsUsername).Find(&t).Error
+	if err != nil {
+		return false
+	}
+	return true
```
alexeykazakov: Check if the error is NotFoundError. If yes then return false. Otherwise return the error. Otherwise if there is some problem with the query or DB different than expected Not Found then you will 1) get an infinite loop when constructing an unique name 2) miss opportunity to catch/log some problem with DB.

tenant/repository.go
```diff
@@ -51,6 +63,15 @@ func (s DBService) GetTenant(tenantID uuid.UUID) (*Tenant, error) {
 	return &t, nil
 }
 
+func (s DBService) NamespaceExists(nsName string) bool {
+	var ns Namespace
+	err := s.db.Table(Namespace{}.TableName()).Where("name = ?", nsName).Find(&ns).Error
+	if err != nil {
+		return false
+	}
+	return true
```
alexeykazakov: Check if the error is NotFoundError. If yes then return false. Otherwise return the error.

controller/tenant.go
```diff
@@ -83,12 +83,15 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
...
 
+	username := tenant.ConstructNsUsername(c.tenantService, env.RetrieveUserName(user.OpenShiftUsername))
+
 	// create openshift config
 	openshiftConfig := openshift.NewConfig(c.config, user.UserData, cluster.User, cluster.Token, cluster.APIURL)
 	tenant := &tenant.Tenant{
 		ID:         ttoken.Subject(),
 		Email:      ttoken.Email(),
 		OSUsername: user.OpenShiftUsername,
+		NsUsername: username,
```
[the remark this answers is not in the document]
MatousJobanek: I called it `ns_username` because there in the templates it is used as: 
```yaml
namespace: ${USER_NAME}-che
``` 
so, I wanted to make it consitent. Unfortunatelly, the fact that we also have OSUsername that is stored in the table as `username` make the stuff more complicated. 

> So, I would like to make it clear across the code that we have clear names for these two concepts here. Username (actual username) and a base-name or prefix which is currently constructed from a username and an index.

The actual `username` is/was used only for the namespaces - there is no other usage of it (if I'm not mistaken). We only use OSUsername. In other words, the `username` is only our artificial thingy (implementation detail) that is retrieved from the real OSUsername. Unfortunately, we used this name of the variable also in the templates - which makes sense in case of user namespace.

So, I'll call it `NsBaseName` and when this PR is merged I'll change also the variables in all templates from `${USER_NAME}` to `${NS_BASE_NAME}` to make it consistent, if you are fine with it.

Why is naming always so hard? :-/

tenant/repository.go
MatousJobanek: Good point - I blindly copied the approach from the [Exists](https://github.com/fabric8-services/fabric8-tenant/blob/master/tenant/repository.go#L33) method that was already there and I didn't realize these cases.
I'll send another PR to fix the `Exists` method as well.

migration/sql-files/006-add-ns-username-column-to-tenant.sql
MatousJobanek: there was a bug in the SQL command - the correct is:
```sql
UPDATE tenants SET ns_base_name = (SELECT name FROM namespaces n WHERE tenants.id = n.tenant_id AND n.type = 'user') 
```
fixed as part of ec19689

controller/tenant.go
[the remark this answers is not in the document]
alexeykazakov: “There are only two hard things in Computer Science: cache invalidation and naming things.”
:)

environment/template.go
```diff
@@ -152,13 +152,11 @@ func (t *Template) ReplaceVars(variables map[string]string) (string, error) {
...
 
-func CollectVars(user, masterUser string, config *configuration.Data) map[string]string {
-	userName := RetrieveUserName(user)
-
+func CollectVars(osUsername, nsBaseName, masterUser string, config *configuration.Data) map[string]string {
 	vars := map[string]string{
-		varUserName:              userName,
-		varProjectUser:           user,
-		varProjectRequestingUser: user,
+		varUserName:              nsBaseName,
```
aslakknutsen: Depending on how this is used but the naming looks odd.. username != basename

aslakknutsen: never mind.. {USER_NAME}-che

PR #676, opened 2018-11-06
[WIP] changes jenkins readiness probes config

hrishin:
Changing Jenkins readiness probes  `initialDelay`
from 40 seconds to 120 seconds.


hrishin:
Let's hold this change for a while and re-evaluate optimal time for `initialDelay`.

Issue #679, opened 2018-11-15
Automate cluster-wide update process


MatousJobanek:
When there is a change introduced to any of the templates, then tenant service should automatically detect this change and update all tenants that are using the old version of the template.

PR #680, opened 2018-11-16
feat(#679): automate cluster-wide update process

MatousJobanek:
For more information about the overall design see [Automatic update of tenant namespaces](https://docs.google.com/document/d/1HKC3A1Zwx8PSfKc01s3Ymgjd0qhLtPDbX0wDUHnt1ng/edit#heading=h.yaz3xk82c7ld)
There is only one difference comparing to the document - a new column in tenants table called `updated_by`. This is to detect which tenant updates failed during the previous update (done by previous tenant service deployment that had different commit sha) and which failed within an ongoing update.

#### Namespace versioning

Originally, the versions weren't properly set/updated, so this necessary step is also part of this PR. In addition, in the cases of jenkins and che namespaces there are used two files (template and resource quotas) for their creation/update, so we need to use versions of both files. In these two cases, the resulting version includes both versions joined by an underscore. eg: `1a2b3c_4d5e6f`.

#### Default che mode

To detect which mode should be used for che (single or multi tenant) we need a user's request context. This would require additional requests to auth as well as to tenant service itself. To avoid it, and to be able to easily trigger an update, the multi-tenant mode has been chosen as the default one. In other words, if the context is nil, then the multi-tenant mode of che is used. The reason is that single tenant is used only by OSIO developers and not by end-users.

Fixes: #679


main.go
```diff
@@ -115,6 +116,9 @@ func main() {
 
 	tenantService := tenant.NewDBService(db)
 
+	// Check & do all tenants update
+	go update.NewTenantsUpdater(10*time.Minute, db, config, authService, clusterService, controller.OSUpdater{}).UpdateAllTenants()
```
alexeykazakov: Maybe worth moving the delay/timeout to configuration?

tenant/repository.go
```diff
@@ -122,6 +124,26 @@ func (s DBService) GetNamespaces(tenantID uuid.UUID) ([]*Namespace, error) {
...
+	for envType, version := range typeWithVersion {
+		conditions = append(conditions, fmt.Sprintf("(n.type = '%s' AND n.version != '%s')", envType, version))
+	}
+	querySuffix := fmt.Sprintf(") AND (n.state != 'failed' OR (n.state = 'failed' AND n.updated_by != '%s')) GROUP BY t.id", commit)
+	finalQuery := baseQuery + strings.Join(conditions, " OR ") + querySuffix
+
+	err := s.db.Raw(finalQuery).Limit(count).Scan(&tenants).Error
+
+	if err != nil {
+		return nil, err
```
alexeykazakov: Or just `return tenants, err`.

tenant/repository_test.go
```diff
@@ -143,6 +146,79 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		// when
+		result, err := svc.GetTenantsToUpdate(testdoubles.GetMappedVersions(environment.DefaultEnvTypes...), 10, "234bcd")
+
+		// then
+		assert.NoError(t, err)
+		assert.Len(t, result, 1)
+		assert.Equal(t, outdated.Tenants[0].ID, result[0].ID)
+	})
+}
+
```
alexeykazakov: Let's also add a test where we have let's say 11 tenants to update. Then GetTenantsToUpdate for the first 5 tenants and check that they are expected ones. Then mark them as "updated" in DB and get the next 5. Check they are expected then "update" and get the next 5 - should get the last one remaining. So, you will imitate the whole cycle.

update/repository.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func NewRepository(tx *sql.Tx) *GormRepository {
+	return &GormRepository{
+		tx: tx,
+	}
+}
+
+func (r *GormRepository) GetStatus() (Status, error) {
+	query := fmt.Sprintf("SELECT status FROM %s", TenantsUpdateTableName)
```
alexeykazakov: Just out of curiosity why you are using a lot of SQL raw stuff instead of GORM. Any particular reason?

update/transaction.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	var tx *sql.Tx
+
+	tx, err = db.Begin()
+	if err != nil {
+		return errors.Wrap(err, "failed to start transaction")
+	}
+
+	err = lockAndDo(tx)
```
alexeykazakov: Let's also add a lock timeout in the beginning of transaction. I guess something like `SET LOCAL lock_timeout = 'XXXs';` where XXX is number of seconds we want to set as the lock timeout. Just to be on the safe side :)

main.go
MatousJobanek: moved here 006ee59

tenant/repository.go
MatousJobanek: done 285ac75

update/repository.go
MatousJobanek: added test here 0d2b231  I covered both cases there - when tenants are successfully updated and when the update failed

tenant/repository_test.go
MatousJobanek: added test here 0d2b231  I covered both cases there - when tenants are successfully updated and when the update failed

update/transaction.go
MatousJobanek: It's part of b0721b0 - I set the timeout to one minute. If you want to ask me why this duration - there is no big reason, maybe just that I didn't want to use something too short nor something too long :-)

update/repository.go
MatousJobanek: as part of the commit, there is a change of the command that starts the execution of integration tests: https://github.com/fabric8-services/fabric8-tenant/commit/b0721b02713d14073c6db94c585e20aa854c874c#diff-552b9cf9c536f6f705ab73d1ab26b005R184
The reason is that some tests were failing only on my machine from the very first moment (on jenkins and on @xcoulon PC the build was green) and the tests were obviously influenced by each other. I took me a very long time to figure out that the problem was parallel test execution (even though that `t.Parallel()` is not used), that was caused by parallel builds - for more information see https://coderwall.com/p/rwgo_a/go-test-package-in-parallel

alexeykazakov: @xcoulon, @MatousJobanek wanna add this `-p 1` https://github.com/fabric8-services/fabric8-tenant/commit/b0721b02713d14073c6db94c585e20aa854c874c#diff-552b9cf9c536f6f705ab73d1ab26b005R184 flag to all our repos (+fabric8-starter if relevant)? :)

tenant/repository_test.go
```diff
@@ -143,6 +146,162 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		controller.Commit = "234bcd"
+		tf.FillDB(t, s.DB, 6, false, "failed", environment.DefaultEnvTypes...)
+
+		svc := tenant.NewDBService(s.DB)
+
+		// when
+		result, err := svc.GetTenantsToUpdate(testdoubles.GetMappedVersions(environment.DefaultEnvTypes...), 10, "234bcd")
+
+		// then
+		assert.NoError(t, err)
```
xcoulon: same here: use `require.NoError`

test/minishift/test_suite.go
```diff
@@ -92,16 +93,21 @@ func prepareConfig(t *testing.T) (*configuration.Data, func()) {
...
 	size := 0
 	for _, objects := range mappedObjects {
 		size += len(objects)
 	}
 	errorChan := make(chan error, size)
 	defer func() {
 		close(errorChan)
+		errWasFound := false
 		for err := range errorChan {
 			assert.NoError(t, err)
```
xcoulon: same here: use `require.NoError`

update/repository_test.go
```diff
@@ -0,0 +1,291 @@
...
+	suite.Run(t, &UpdateRepoTestSuite{DBTestSuite: gormsupport.NewDBTestSuite("../config.yaml")})
+}
+
+func (s *UpdateRepoTestSuite) TestUpdateAndGetStatus() {
+	s.T().Run("set and get status should pass", func(t *testing.T) {
+		// given
+		err := update.Transaction(s.DB, func(tx *gorm.DB) error {
+			return update.NewRepository(tx).UpdateStatus(update.Updating)
+		})
+		require.NoError(s.T(), err)
```
xcoulon: use `require.NoError(t, err)` otherwise the test execution will crash with a `panic` in case of error

update/repository_test.go
```diff
@@ -0,0 +1,291 @@
...
+		})
+
+		// then
+		assert.NoError(t, err)
+		var tenantsUpdate *update.TenantsUpdate
+		err = update.Transaction(s.DB, func(tx *gorm.DB) error {
+			tenantsUpdate, err = update.NewRepository(tx).GetTenantsUpdate()
+			return err
+		})
+		assert.NoError(t, err)
```
xcoulon: same here: use `require.NoError`

update/repository_test.go
```diff
@@ -0,0 +1,291 @@
...
+			}
+			if err := repo.UpdateStatus(update.Failed); err != nil {
+				return err
+			}
+			if err := repo.IncrementFailedCount(); err != nil {
+				return err
+			}
+			return updateVersionsTo(repo, "000abc")
+		})
+		require.NoError(s.T(), err)
```
xcoulon: `require.NoError(t, ..)`

controller/tenant.go
```diff
@@ -177,27 +177,85 @@ func (c *TenantController) Update(ctx *app.UpdateTenantContext) error {
...
+		}
+	}
+	return nil
+}
+
+type UpdateExecutor interface {
+	Update(ctx context.Context, tenantService tenant.Service, openshiftConfig openshift.Config, t *tenant.Tenant, envTypes []string) (map[string]string, error)
+}
+
+type OSUpdater struct {
```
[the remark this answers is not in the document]
MatousJobanek: makes sense, renamed in ce02e11

Makefile
```diff
@@ -32,7 +32,7 @@ export GIT_COMMITTER_EMAIL
 COMMIT=$(shell git rev-parse HEAD)
 GITUNTRACKEDCHANGES := $(shell git status --porcelain --untracked-files=no)
 ifneq ($(GITUNTRACKEDCHANGES),)
-COMMIT := $(COMMIT)-dirty
+COMMIT := $(COMMIT)
```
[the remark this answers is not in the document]
MatousJobanek: yes, you are right :-) 3ef7fbb

environment/service_test.go
```diff
@@ -61,45 +61,47 @@ func TestGetAllTemplatesForAllTypes(t *testing.T) {
...
-				assert.Equal(t, "567efg", environment.GetLabelVersion(objects[0]))
-				assert.Equal(t, "yxw987", environment.GetLabel(objects[0], environment.FieldVersionQuotas))
-			} else {
-				if strings.Contains(env.Templates[0].Filename, "mt") {
-					assert.Equal(t, "234bcd", environment.GetLabelVersion(objects[0]))
-					assert.Equal(t, "zyx098", environment.GetLabel(objects[0], environment.FieldVersionQuotas))
+			// then
+			require.NoError(t, err)
+			assert.Equal(t, env.Name, envType)
+			if envType == "che" || envType == "jenkins" {
```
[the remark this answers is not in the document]
MatousJobanek: thanks for this proposal - it looks much better: 8d54416

environment/template.go
```diff
@@ -85,29 +85,31 @@ type Template struct {
...
+	Version       string
 }
 
 var (
 	specialCharRegexp = regexp.MustCompile("[^a-z0-9]")
 	variableRegexp    = regexp.MustCompile(`\${([A-Z_0-9]+)}`)
 )
 
-func newTemplate(filename string, defaultParams map[string]string) *Template {
+func newTemplate(filename string, defaultParams map[string]string, version string) *Template {
```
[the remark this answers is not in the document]
MatousJobanek: I use the pointer later on - when the content of the template is set. And it is easier to create the pointer directly in the constructor then in a loop by creating a copy of the value for the whole array. And since there is only one state of each template per one request (see that the instances of templates are [created](https://github.com/fabric8-services/fabric8-tenant/pull/680/files#diff-548bfeebae84ae77e57356c9ba042b4aR40) on demand - by function and not as variable), then I don't see any point in removing the pointer in the constructor

test/minishift/test_suite.go
MatousJobanek: check the parameter `required` ;-). if this evaluation is part of `// given` section, then if there was any error found, then the setup is stopped: https://github.com/fabric8-services/fabric8-tenant/pull/680/files/ff391be402df912216ddd888fc1e9eb023fd3bee#diff-e526609efd4c9f26c792830751f918e4R110 as a benefit we will get the context of all errors that occured during the setup.

tenant/repository_test.go
```diff
@@ -143,6 +146,162 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		testdoubles.SetTemplateVersions()
+		fxt := tf.FillDB(t, s.DB, 11, false, "ready", environment.DefaultEnvTypes...)
+		svc := tenant.NewDBService(s.DB)
+		mappedVersions := testdoubles.GetMappedVersions(environment.DefaultEnvTypes...)
+
+		// when
+		firstBatch, err := svc.GetTenantsToUpdate(mappedVersions, 5, "xyz")
+
+		// then
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
MatousJobanek: a64525f

environment/template.go
```diff
@@ -186,13 +188,13 @@ func getVariables(config *configuration.Data) map[string]string {
 	return templateVars
 }
 
-func merge(target, second map[string]string) map[string]string {
+func merge(target, second map[string]string, overwrite bool) map[string]string {
```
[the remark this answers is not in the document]
xcoulon: ok

tenant/repository_test.go
```diff
@@ -143,6 +146,162 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		assert.Len(t, firstBatch, 5)
+		assertContentOfTenants(t, firstBatch, fxt.Tenants, true)
+		updateAllTenants(t, firstBatch, svc, false)
+
+		// and when
+		secondBatch, err := svc.GetTenantsToUpdate(mappedVersions, 5, "xyz")
+
+		// then
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
MatousJobanek: a64525f

tenant/repository_test.go
```diff
@@ -143,6 +146,162 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		assert.Len(t, secondBatch, 5)
+		assertContentOfTenants(t, secondBatch, fxt.Tenants, true)
+		assertContentOfTenants(t, secondBatch, firstBatch, false)
+		updateAllTenants(t, secondBatch, svc, true)
+
+		// and when
+		thirdBatch, err := svc.GetTenantsToUpdate(mappedVersions, 5, "xyz")
+
+		// then
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
MatousJobanek: a64525f

tenant/repository_test.go
MatousJobanek: lazy assertion

update/repository_test.go
MatousJobanek: wrong copy-paste :-) a64525f

update/repository_test.go
MatousJobanek: a64525f

update/repository_test.go
MatousJobanek: wrong copy-paste :-) a64525f

environment/template.go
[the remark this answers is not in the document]
xcoulon: it's just that in general, you should only return a pointer if the element needs to be changed by the caller but the variable belongs to the callee. Same with func arguments: only use a pointer if fields needs to be changed, otherwise, pass a copy.

tenant/repository_test.go
```diff
@@ -143,6 +146,162 @@ func (s *TenantServiceTestSuite) TestLookupTenantByNamespace() {
...
+		assert.Len(t, lastBatch, 0)
+	})
+}
+
+func updateAllTenants(t *testing.T, toUpdate []*tenant.Tenant, svc tenant.Service, failed bool) {
+	mappedVersions := testdoubles.GetMappedVersions(environment.DefaultEnvTypes...)
+	for _, tnnt := range toUpdate {
+		namespaces, err := svc.GetNamespaces(tnnt.ID)
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
xcoulon: hum, that's not my understanding of `assert.NoError()`: it records an error but continues the execution of the code, whereas `require.NoError` stops if the condition is not met/validated.

test/minishift/test_suite.go
xcoulon: so why don't you just replace line 106 with something like:
```
if required {
  require.NoError(t, err)
}
``` 
and otherwise, just ignore the error (or log it, if needed)

tenant/repository_test.go
xcoulon: sorry, I'm not following you

environment/template.go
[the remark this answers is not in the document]
MatousJobanek: I think that the topic of pointers/values is more complicated. But this is a different story for sure. 
I can do this 6fc3ff6 if you think that it looks better/is cleaner

tenant/repository_test.go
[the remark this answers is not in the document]
MatousJobanek: sorry for not being clear with my explanation - I wasn't talking about `assert.NoError()` but about the test in general. The test tries to get namespaces and then loops through all namespaces and performs additional assertions for each namespace. It means that if there is no namespace returned, then no additional assertions are performed.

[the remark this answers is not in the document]
xcoulon: hum, but then shouldn't you do something like 
```
if err != nil {
  continue
}
```
to pass to the next element in the `toUpdate` slice?

environment/template.go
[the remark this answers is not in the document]
xcoulon: yes, but that's a rule that I try to use when I need to decide whether to use pointers or not. Also, golang is very fast at copying memory blocks, so passing a copy isn't so expensive.
I'd rather not use pointers when really needed, actualy

tenant/repository_test.go
MatousJobanek: sorry, I didn't want to write lazy, but soft assertion :-).
I just like the concept when you let all assertions evaluate and then get all failure messages together.

environment/template.go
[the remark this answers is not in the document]
MatousJobanek: Anyway, this is exactly the case when I need to change the instance of the template - to set the content and potentially change the version. So, are you fine with the proposed solution?

tenant/repository_test.go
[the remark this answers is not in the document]
MatousJobanek: now I don't follow - do you mean something like this?:
```go
    for _, tnnt := range toUpdate {
		namespaces, err := svc.GetNamespaces(tnnt.ID)
		assert.NoError(t, err)
		if err != nil {
			continue
		}
		for _, ns := range namespaces {
			if failed {
				ns.State = "failed"
			} else {
				ns.Version = mappedVersions[string(ns.Type)]
				ns.State = "ready"
			}
			ns.UpdatedBy = "xyz"
			assert.NoError(t, svc.SaveNamespace(ns))
		}
	}
```
if so, then isn't the result same, just with additional three lines of the code?

test/minishift/test_suite.go
xcoulon: ok, got it.

Issue #684, opened 2018-11-26
Add integration tests that would use Minishift


MatousJobanek:
Add integration tests that would use Minishift for creation/update/deletion of the namespaces.
Enable it only for local testing as it is not guaranteed that we would be able to run Minishift in ci-centos environment.

PR #685, opened 2018-11-26
 chore(#684) introduce tests that use Minishift for creation/update/deletion of tenant namespaces

MatousJobanek:
This PR adds one bigger integration test (executed only locally) that runs happy path of creation -> update -> cleanup -> deletion of all tenant namespaces for a single user on Minishift.
Apart from that, it brings functions that will help with adding more integration tests in the future (eg. automated update) and with checking any possible regressions (eg. big core refactoring PR).

Depends on: #683
Fixes: #684


auth/service.go
```diff
@@ -22,26 +22,45 @@ import (
...
-	ClientOptions []configuration.HTTPClientOption
-	SaToken       string
+type Service interface {
+	GetUser(ctx context.Context) (*User, error)
+	GetAuthURL() string
+	NewSaClient() (*authclient.Client, error)
+	ResolveUserToken(ctx context.Context, target, userToken string) (user, accessToken string, err error)
+	ResolveSaToken(ctx context.Context, target string) (username, accessToken string, err error)
+	GetClientOptions() []configuration.HTTPClientOption
+	GetPublicKeys() ([]*rsa.PublicKey, error)
```
alexeykazakov: We already have Auth service in common. We need to add more stuff like this to common and reuse it here.
It's related to https://github.com/fabric8-services/fabric8-cluster/issues/8 (@dipak-pawar wink wink)

auth/service.go
```diff
@@ -142,33 +165,33 @@ func (s *Service) getOAuthToken(ctx context.Context) (*string, error) {
 
 	validationerror := ValidateResponse(ctx, client, res)
```
xcoulon: variable here could be simply named `err` (not a required change)

cluster/service.go
```diff
@@ -54,7 +54,7 @@ type clusterService struct {
 }
 
 // NewClusterService creates an instance of service that using the Auth service retrieves information about clusters
-func NewClusterService(refreshInt time.Duration, authService *auth.Service) Service {
+func NewClusterService(refreshInt time.Duration, authService auth.Service) Service {
```
xcoulon: why don't we pass the client options in the constructor here. rather than using `s.authService.GetClientOptions()...` on line 156?

controller/tenant_minishift_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// when setup is called
+	goatest.SetupTenantAccepted(s.T(), createUserContext(s.T(), id.String()), svc, ctrl)
+
+	// then
+	repo := tenant.NewDBService(s.DB)
+	tnnt, err := repo.GetTenant(id)
+	assert.NoError(s.T(), err)
+
+	iteration := 0
```
xcoulon: are you trying to get the namespaces with a certain amount of time ? if so, you may want to use channels with a time out instead. You find some example in https://github.com/fabric8-services/fabric8-cluster/blob/master/application/transaction/transaction.go#L53-L84

controller/tenant_minishift_test.go
```diff
@@ -0,0 +1,115 @@
...
+	s.T().Run("update namespaces", func(t *testing.T) {
+		// given
+		testdoubles.SetTemplateSameVersion("2abcd")
+
+		// when update is called
+		goatest.UpdateTenantAccepted(t, createUserContext(t, id.String()), svc, ctrl)
+
+		// then
+		namespaces, err := tenant.NewDBService(s.DB).GetNamespaces(id)
+		assert.NoError(t, err)
```
xcoulon: you probably need a `require.NoError` here to stop the test in case of error

controller/tenant_minishift_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+
+	s.T().Run("only clean namespaces", func(t *testing.T) {
+
+		// when clean is called
+		goatest.CleanTenantNoContent(t, createUserContext(t, id.String()), svc, ctrl, false)
+
+		// then
+		namespaces, err := tenant.NewDBService(s.DB).GetNamespaces(id)
+		assert.NoError(t, err)
```
xcoulon: same here with `require.NoError` instead of `assert.NoError`

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+	minishiftUrl        string
+	minishiftAdminName  string
+	minishiftAdminToken string
+	minishiftUserName   string
+	minishiftUserToken  string
+}
+
+func (s *TestSuite) SetupTest() {
+	resource.Require(s.T(), resource.Database, "F8_MINISHIFT")
+	s.minishiftUrl = os.Getenv("F8_MINISHIFT_URL")
```
xcoulon: this could be handled by a "minishift configuration" type managed by spf13/viper (similar to the regular service configuration)

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+	s.minishiftAdminName = os.Getenv("F8_MINISHIFT_ADMIN_NAME")
+	s.minishiftAdminToken = os.Getenv("F8_MINISHIFT_ADMIN_TOKEN")
+	s.minishiftUserName = os.Getenv("F8_MINISHIFT_USER_NAME")
+	s.minishiftUserToken = os.Getenv("F8_MINISHIFT_USER_TOKEN")
+	s.DBTestSuite.SetupTest()
+	s.Config, s.toReset = prepareConfig(s.T())
+
+	log.InitializeLogger(s.Config.IsLogJSON(), s.Config.GetLogLevel())
+
+	s.ClusterService = &stub.ClusterService{
```
xcoulon: this could be initialized in the `SetupSuite` function, once for all

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func VerifyObjectsPresence(t *testing.T, mappedObjects map[string]environment.Objects, options openshift.ApplyOptions, version string) {
+	for ns, objects := range mappedObjects {
+		var wg sync.WaitGroup
+		for _, obj := range objects {
+			wg.Add(1)
+			go func(toMarkAsDone *sync.WaitGroup, t *testing.T) {
+				iterations := 0
+				defer toMarkAsDone.Done()
```
xcoulon: I believe you can use the `wg` variable directly from here. No need to pass it as an arg named `toMarkAsDone`

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+				for {
+					if openshift.IsOfKind(environment.ValKindProjectRequest, environment.ValKindProject, environment.ValKindNamespace)(obj) {
+						if environment.GetKind(obj) == environment.ValKindProjectRequest {
+							obj["kind"] = environment.ValKindNamespace
+						}
+						response, err := openshift.Apply(obj, "GET", options)
+						if err == nil && environment.HasValidStatus(response) {
+							break
+						} else if iterations >= 20 {
+							assert.NoError(t, err)
```
xcoulon: If I understand correctly, you're waiting for up to 20 iterations and if there's still an error, then the test will fail ? As suggested above, I believe that the code would be easier to read if you use a `select channels` with a timeout, instead.

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+		MasterUser: s.minishiftAdminName,
+		Token:      s.minishiftAdminToken,
+	}
+
+	templs, err := openshift.LoadProcessedTemplates(context.Background(), config, s.minishiftUserName, nsBaseName)
+	assert.NoError(s.T(), err)
```
xcoulon: use `require.NoError` here as well

test/minishift/test_suite.go
```diff
@@ -0,0 +1,161 @@
...
+	mapped, err := openshift.MapByNamespaceAndSort(templs)
+	assert.NoError(s.T(), err)
```
xcoulon: use `require.NoError` here too, please

test/stub/service.go
```diff
@@ -0,0 +1,83 @@
+package stub
```
xcoulon: if this is just for testing, you can also use `minimock` to generate mocks which you can use in your test (and you can customize the functions for each test as well).
See:
https://github.com/fabric8-services/fabric8-auth/blob/master/Makefile#L148-L160 + https://github.com/fabric8-services/fabric8-auth/blob/master/Gopkg.toml#L49 for the setup and code generation, then https://github.com/fabric8-services/fabric8-auth/blob/master/test/witservice.go#L14-L30 for using the mocks

controller/tenant_minishift_test.go
MatousJobanek: I think that the main difference is that I'm not running the iteration in goroutine.

controller/tenant_minishift_test.go
MatousJobanek: I'm not sure - I like the lazy assertion behavior. I use require only in the `//given` part of the test to stop it if there is an error in the setup.

test/minishift/test_suite.go
MatousJobanek: well, I was too lazy to create a special config implementation just because of a few variables that are used only in tests anyway. But I can do that :-)

test/minishift/test_suite.go
MatousJobanek: I want to keep it consistent - use the same approach for both ClusterServices and AuthService

test/minishift/test_suite.go
MatousJobanek: I'm afraid that not - at least I faced an issue with it some time ago. In addition, I need to use a pointer to it otherwise it would use just a copy of the value.

test/minishift/test_suite.go
MatousJobanek: I'm not sure if `channels` is not an overkill:
* I don't need to store all errors - it is enough just the last one
* I don't need to distinguish between timeout and error
* the likelihood that the error would occur is low - at least I haven't received any, I check the error just for sure
* the most important check is `environment.HasValidStatus(response)`

test/stub/service.go
MatousJobanek: > if this is just for testing, you can also use minimock to generate mocks which you can use in your test (and you can customize the functions for each test as well).

for now, it is only for tests. In the future, I have a plan to use it for running tenant service locally.

auth/service.go
MatousJobanek: well, not changed in this PR, but why not... 
renamed in 731ab69

cluster/service.go
MatousJobanek: because this information is stored in auth service and I'm passing the auth service to the constructor anyway. 
But I see your point - this didn't look so bad in the previous solution - without the interface. Now it looks a bit weird. 
changed here: 3d217d2 
thanks for pointing it out

controller/tenant_minishift_test.go
xcoulon: but if you use `assert`, then the test will continue. Is it what you really want?

controller/tenant_minishift_test.go
xcoulon: yes, indeed.

controller/tenant_minishift_test.go
MatousJobanek: well, yes. in that case, it also evaluates the other assertions which can give me more information on what has failed or what is wrong

test/minishift/test_suite.go
MatousJobanek: hm... let me try that :-)

test/minishift/test_suite.go
MatousJobanek: I'll try to improve it ;-)

controller/tenant_minishift_test.go
xcoulon: my raw idea was to use 2 channels. One for the timeout, and another one that emits a message when the namespaces are created (with a loop to check at regular intervals - this needs to stay). But the function to test the availability of the channels could be externalized. does it make sense?

controller/tenant_minishift_test.go
xcoulon: ok

controller/tenant_minishift_test.go
MatousJobanek: this should be better I think - I hope that it helps: 0093df3

test/minishift/test_suite.go
MatousJobanek: improved as part of 0093df3 
see here https://github.com/fabric8-services/fabric8-tenant/pull/685/commits/0093df3ccbb9f9bc8a1d000ea6aec5510bab03c4#diff-e526609efd4c9f26c792830751f918e4R103

test/minishift/test_suite.go
MatousJobanek: Yes, you are right - fixed here ebc661b

test/minishift/test_suite.go
MatousJobanek: done in ebdd969

test/awaitility.go
```diff
@@ -0,0 +1,25 @@
...
+func (a Awaitility) Until(do func() error) error {
+	err := do()
+	if err == nil {
+		return nil
+	} else if a.before.Before(time.Now().Add(-a.timeout)) {
+		return err
+	}
+	time.Sleep(500 * time.Millisecond)
+	return a.Until(do)
+}
```
nurali-techie: go way of doing this - https://play.golang.org/p/TJS0qxJElrj
less logic and more magic ;-)

xcoulon: yes, this was my initial thought as well. Thanks @nurali-techie for the simplified example!

nurali-techie: now `before` is not used anywhere.  I guess .. no need of struct and can be done with func only.

MatousJobanek: Yes, this is what happens when I'm doing several things on multiple branches at once :-( 
removed in 61fe716 
Thanks @nurali-techie

test/minishift/test_suite.go
```diff
@@ -0,0 +1,172 @@
...
+	minishiftConfig *Data
+}
+
+func (s *TestSuite) SetupTest() {
+	resource.Require(s.T(), resource.Database)
+
+	config, err := NewData()
+	require.NoError(s.T(), err)
+	if config.GetMinishiftURL() == "" {
+		s.T().Skipf("Minishift URL using a variable [%s] has to be set", varMinishiftURL)
```
dipak-pawar: This looks confusing to me. because you have to set `MINISHIFT_URL` not `minishift.url`. maybe rewording is enough.(Do we want to set ENV or something else)

test/awaitility.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (a Awaitility) Until(do func() error) error {
+	ctx, cancel := context.WithTimeout(context.Background(), a.timeout)
+	defer cancel()
+	var err error
+	for {
+		select {
+		case <-ctx.Done():
+			return err
+		case <-time.After(500 * time.Millisecond):
```
dipak-pawar: Also we are talking to openshift master when this is called, I am not sure will it be changed(some resources inside namespace at every iteration) in every 0.5 second. Maybe increase it `1 or 2 s`

test/minishift/test_suite.go
```diff
@@ -0,0 +1,172 @@
...
+	for ns, objects := range mappedObjects {
+		for _, obj := range objects {
+			wg.Add(1)
+			go func(obj environment.Object, ns string) {
+				defer wg.Done()
+				errorChan <- test.WaitWithTimeout(1 * time.Minute).Until(objectIsUpToDate(obj, ns, options, version))
+			}(obj, ns)
+		}
+	}
+	wg.Wait()
```
dipak-pawar: maybe worth to use timeout instead of waiting on wg done, otherwise we'll have a deadlock

test/stub/service.go
```diff
@@ -0,0 +1,79 @@
...
+func (s *ClusterService) GetClusters(ctx context.Context) []cluster.Cluster {
+	cl, _ := s.GetCluster(ctx, "")
+	return []cluster.Cluster{cl}
+
+}
+
+func (s *ClusterService) Stop() {
+}
+
+type AuthService struct {
```
dipak-pawar: can't we introduce minimock for AuthService? It will help us to modify this stub each time for every single change of `AuthService`.

test/stub/service.go
dipak-pawar: @MatousJobanek I understand your point. but does it mean we'll be removing this stub and asking everyone to run cluster locally before running this integration tests. I think we should introduce `minimock` to avoid more maintenance in future

.make/test.mk
```diff
@@ -117,6 +117,11 @@ ALL_PKGS_EXCLUDE_PATTERN = 'vendor\|app\|tool\/cli\|design\|client\|test'
 GOANALYSIS_PKGS_EXCLUDE_PATTERN="vendor|app|client|tool/cli"
 GOANALYSIS_DIRS=$(shell go list -f {{.Dir}} ./... | grep -v -E $(GOANALYSIS_PKGS_EXCLUDE_PATTERN))
 
+MINISHIFT_USER_NAME ?= developer$(shell date +'%H.%M.%S')
```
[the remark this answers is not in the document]
MatousJobanek: If you run the test suite twice without any enough time gap between the test executions, then you will face the problem that a namespace is in `Terminating` state which can take some time and will cause failures in the test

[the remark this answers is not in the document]
MatousJobanek: ...This is why I use a new user for each run - to have different namespace names ;-)

.make/test.mk
```diff
@@ -117,6 +117,11 @@ ALL_PKGS_EXCLUDE_PATTERN = 'vendor\|app\|tool\/cli\|design\|client\|test'
...
+MINISHIFT_USER_TOKEN ?= $(shell oc login -u=$(MINISHIFT_USER_NAME) -p=developer > /dev/null && oc whoami -t)
+MINISHIFT_ADMIN_NAME ?= admin
```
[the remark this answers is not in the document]
MatousJobanek: I don't create a new `admin` because it is not needed - the admin is used only for having `cluster-admin` rights when creating/updating/deleting the namespaces.

.make/test.mk
```diff
@@ -200,6 +205,16 @@ test-remote-no-coverage: prebuild-check $(SOURCES)
...
+.PHONY: test-with-minishift
+## Runs the tests which require availability of minishift as well as DB.
+test-with-minishift: prebuild-check migrate-database
+	$(call log-info,"Running test: $@")
+	$(eval TEST_PACKAGES:=$(shell go list ./... | grep -v $(ALL_PKGS_EXCLUDE_PATTERN)))
+	F8_DEVELOPER_MODE_ENABLED=1 F8_RESOURCE_DATABASE=1 F8_RESOURCE_UNIT_TEST=0 F8_POSTGRES_DATABASE=postgres \
+	F8_MINISHIFT_USER_NAME=$(MINISHIFT_USER_NAME) F8_MINISHIFT_USER_TOKEN=$(MINISHIFT_USER_TOKEN) F8_MINISHIFT_URL=$(MINISHIFT_URL) \
+	F8_MINISHIFT_ADMIN_NAME=$(MINISHIFT_ADMIN_NAME) F8_MINISHIFT_ADMIN_TOKEN=$(MINISHIFT_ADMIN_TOKEN) \
+	go test -v -run ".*Minishift.*" $(TEST_PACKAGES)
+
```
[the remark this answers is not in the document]
MatousJobanek: I wouldn't do that - in case you would need to rerun the test suite using the same user and token

test/awaitility.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (a Awaitility) Until(do func() error) error {
+	ctx, cancel := context.WithTimeout(context.Background(), a.timeout)
+	defer cancel()
+	var err error
+	for {
+		select {
+		case <-ctx.Done():
+			return err
```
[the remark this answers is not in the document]
MatousJobanek: Not at all, we need to return the error that was retrieved from getting/checking the namespace existence - if there is a timeout. If the timeout is reached and the err (for the last check) is nil, then it is fine that the test should pass - it means that namespace/object is ready ;-)

test/minishift/test_suite.go
MatousJobanek: well, this is getting to be complicated - go don't even have `MINISHIFT_URL` but `F8_MINISHIFT_URL`. I'll reword it not to use the variable name - I'll use:
`Test is skipped because of missing Minishift variables`

test/stub/service.go
MatousJobanek: I don't follow @dipak-pawar

> but does it mean we'll be removing this stub

why removing this stub?

> asking everyone to run cluster locally before running this integration tests.

you need to run minishift to make the test running for sure - this is the intend of the test

auth/service.go
```diff
@@ -140,35 +158,35 @@ func (s *Service) getOAuthToken(ctx context.Context) (*string, error) {
...
 	}
 	token, err := client.DecodeOauthToken(res)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error from server %q", s.Config.GetAuthURL())
+		return nil, errors.Wrapf(err, "error from server %q", s.config.GetAuthURL())
 	}
 
 	if token.AccessToken == nil || *token.AccessToken == "" {
-		return nil, fmt.Errorf("received empty token from server %q", s.Config.GetAuthURL())
+		return nil, fmt.Errorf("received empty token from server %q", s.config.GetAuthURL())
```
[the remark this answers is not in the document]
MatousJobanek: Please feel free to create an issue for it - this is not the goal of this PR.
Btw, if you remember, we discussed it in our week-sync meeting already ;-)

test/minishift/test_suite.go
xcoulon: agreed, I don't see why there would be a deadlock here.

test/minishift/test_suite.go
xcoulon: sounds good enough

test/awaitility.go
xcoulon: @MatousJobanek if an event comes from the `ctx.Done()` channel, then it doesn't really matter how much time you'll wait, does it ? My point is, giving `1s` or `2s` of timeout just gives a little more time for the operation to complete, but it's not going to block the test for the entire `timeout` duration.

README.adoc
```diff
@@ -106,6 +106,24 @@ To stop DB:
...
+https://docs.okd.io/latest/minishift/using/addons.html#installing-addons[Install] and https://docs.okd.io/latest/minishift/using/addons.html#apply-addons[apply] the https://docs.okd.io/latest/minishift/using/addons.html#default-addons[admin-user] Minishift addon
+```
+$ minishift addons apply admin-user
+```
+
+Run Minishift integration tests. Some of the tests may take a longer time to finish
+```
+$ make test-with-minishift
+```
+
```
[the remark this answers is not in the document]
MatousJobanek: added note 175a173

auth/service.go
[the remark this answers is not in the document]
xcoulon: agreed, this can be addressed in a separate issue/PR

test/minishift/test_suite.go
MatousJobanek: changed in 91b9d52

.make/test.mk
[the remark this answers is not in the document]
dipak-pawar: ok but then we'll have n number of users (n = number of iterations) in the long run.

.make/test.mk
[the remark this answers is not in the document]
dipak-pawar: thanks for explaination

.make/test.mk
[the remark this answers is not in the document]
dipak-pawar: ok but then we'll have n number of users (n = number of iterations tests run) in the long run.

auth/service.go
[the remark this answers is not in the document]
dipak-pawar: as far as I remember it's moving some code in common. I am pointing here we have some functionality already in common. but anyways let's address it in separate PR.

test/minishift/test_suite.go
dipak-pawar: Hmm I missed it. Thanks for clarification

test/awaitility.go
MatousJobanek: well, currently I'm running these checks in goroutine, which means that changing it to `1s` doesn't influence the resulting time significantly. Originally it was different - it was executed in for cycle sequentially and if you multiply the number of checks by half a sec, then you will get a quite big time difference. Then, I have one test that runs this wait logic in sequence 12-times - this is the difference of 6 seconds. 
I'm willing to change it to `1s` but not `2s`. Still, my main question is - what is the benefit of changing it?

test/stub/service.go
dipak-pawar: ok

.make/test.mk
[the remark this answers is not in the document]
MatousJobanek: > ok but then we'll have n number of users (n = number of iterations tests run) in the long run.

I don't follow - the single user is used for all tests in the test-suite execution. which means that a new user per whole test-suite execution... 

anyway, if you have any idea how to:
* keep the possibility of rerunning the test suite using the same user
* do not pollute minishift with new users
* avoid conflicts with the terminating state

then I'm open to change it

test/awaitility.go
dipak-pawar: I am not sure if we get some object state change with every `0.5s`.

auth/service.go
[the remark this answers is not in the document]
MatousJobanek: the main point was to merge my code and the code in common. Then put the resulting code in common and when we have the complete functionality that won't be changed, then I'll replace it in tenant service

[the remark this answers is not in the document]
xcoulon: yes, let's not block this PR for some changes that could go in `common` afterwards.

test/awaitility.go
MatousJobanek: changed to `1s` 6a41e1b

.make/test.mk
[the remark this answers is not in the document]
dipak-pawar: how about having some make target to clean all users created by tests. This way anyone can use and clean it up if needed.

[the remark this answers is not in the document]
MatousJobanek: btw, if the test (or any of the future minishift tests such as this one https://github.com/MatousJobanek/fabric8-tenant/blob/automated-update-with-minishift/update/update_minishift_test.go) fails (for any reason) without the proper clean-up of all namespaces then you will have a much bigger problem with a huge number of existing namespaces than just having a few additional users :-)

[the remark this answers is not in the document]
MatousJobanek: > how about having some make target to clean all users created by tests. This way anyone can use and clean it up if needed.

good idea - let me try to write something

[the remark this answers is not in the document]
MatousJobanek: @dipak-pawar is this something that you expected? 9350605

Issue #688, opened 2018-11-30
Print error response message retrieved from auth


MatousJobanek:
When an error is returned from auth service https://github.com/fabric8-services/fabric8-auth/blob/master/controller/token.go#L289 then read also the response message from header https://github.com/fabric8-services/fabric8-auth/blob/master/controller/token.go#L206 and log it apart from the error itself as well https://github.com/fabric8-services/fabric8-tenant/blob/master/auth/service.go#L204-L207

It is needed to get more info about `Unauthorized error` https://github.com/openshiftio/openshift.io/issues/4601

PR #689, opened 2018-11-30
fix(#688): add error msg when token retrieval failed

MatousJobanek:
When token retrieval fails, then add also error message from header to the resulting error

Fix: #688

### DIFF
--- a/auth/service.go
+++ b/auth/service.go
@@ -203,7 +203,7 @@ func (s *authService) ResolveTargetToken(ctx context.Context, target, token stri
 	}
 	res, err := client.RetrieveToken(ctx, authclient.RetrieveTokenPath(), target, &forcePull)
 	if err != nil {
-		return "", "", errors.Wrapf(err, "error while resolving the token for %s", target)
+		return "", "", constructResolveTokenError(err, res, target)
 	}
 	defer func() {
 		ioutil.ReadAll(res.Body)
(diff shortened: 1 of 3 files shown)


auth/service.go
```diff
@@ -255,6 +255,15 @@ func (s *authService) GetAuthUserData(ctx context.Context, tenantToken TenantTok
 	return user.Data.Attributes, nil
 }
 
+func constructResolveTokenError(err error, res *http.Response, target string) error {
+	if res != nil && res.Header != nil {
+		if msg := res.Header["WWW-Authenticate"]; len(msg) > 0 {
```
xcoulon: you could probably write something like: 
```
if msg, ok := res.Header["WWW-Authenticate"]; ok {
   ...
}
```

xcoulon: also, can you write a "whitebox" test for this function, so you can test both cases (ie, an HTTP response from `auth` with and without the `WWW-Authenticate` header) ?

MatousJobanek: if statement changed 131df61 

as for the whitebox test - this case is already covered by all tests that verify message of the format "error while resolving the token for" - eg this one: https://github.com/fabric8-services/fabric8-tenant/pull/689/files/588b0887f60d59fd269e4becc93f5621d507f112#diff-c18bfd11fd5d373e7c777c050f0bcf9eR44

xcoulon: yes, I've seen this test, but you did not cover the case where the response does not contain the `WWW-Authenticate` header, which is why I suggested this "whitebox test". But if you add another test with a response without the aforementioned header, it'll be fine, too ;) As you like.

xcoulon: ok, thanks! (sorry, I missed that in the existing code)

PR #700, opened 2018-12-13
create environment.Type and use it instead of NamespaceType

MatousJobanek:
as part of an attempt to lower the number of changes in the huge refactoring [PR](https://github.com/fabric8-services/fabric8-tenant/pull/673), I'm bringing one smaller part as a separated PR.

This one replaces `NamespaceType` by `environment.Type` and uses it in all cases where the code refers to the type of environment/namespace

### DIFF
--- a/environment/service.go
+++ b/environment/service.go
@@ -30,20 +30,20 @@ var (
 	VersionFabric8TenantJenkinsQuotasFile string
 	VersionFabric8TenantCheQuotasFile     string
 	VersionFabric8TenantDeployFile        string
-	DefaultEnvTypes                       = []string{"che", "jenkins", "user", "run", "stage"}
+	DefaultEnvTypes                       = []Type{TypeChe, TypeJenkins, TypeRun, TypeStage, TypeUser}
 )
 
 type Templates []*Template
 
-func RetrieveMappedTemplates() map[string]Templates {
-	return map[string]Templates{
-		"run":   tmpl(deploy("run"), "fabric8-tenant-deploy.yml"),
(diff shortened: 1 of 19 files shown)


environment/service.go
```diff
@@ -30,20 +30,20 @@ var (
 	VersionFabric8TenantJenkinsQuotasFile string
 	VersionFabric8TenantCheQuotasFile     string
 	VersionFabric8TenantDeployFile        string
-	DefaultEnvTypes                       = []string{"che", "jenkins", "user", "run", "stage"}
+	DefaultEnvTypes                       = []Type{TypeChe, TypeJenkins, TypeRun, TypeStage, TypeUser}
```
alexeykazakov: I guess it doesn't matter but you are changing the order of types here. Just wanted to bring your attention in case we actually relay on the order somewhere. If not then it's fine.

MatousJobanek: I'm aware of it and it doesn't influence anything - nothing in the code relies on the order of the values in this slice

tenant/tenant.go
```diff
@@ -86,26 +49,26 @@ func (m Namespace) TableName() string {
...
 // GetNamespaceType attempts to extract the namespace type based on namespace name
-func GetNamespaceType(name, nsBaseName string) NamespaceType {
+func GetNamespaceType(name, nsBaseName string) environment.Type {
 	if name == nsBaseName {
-		return TypeUser
+		return environment.TypeUser
 	}
 	if strings.HasSuffix(name, "-jenkins") {
-		return TypeJenkins
+		return environment.TypeJenkins
```
dipak-pawar: maybe we can switch case here?

PR #704, opened 2018-12-14
change reading of F8_AUTOMATED_UPDATE_ENABLED var to be set as param

MatousJobanek:
https://github.com/openshiftio/saas-openshiftio/pull/1170

### DIFF
--- a/OpenShiftTemplate.yml
+++ b/OpenShiftTemplate.yml
@@ -133,10 +133,7 @@ objects:
                 name: f8tenant
                 key: environment
           - name: F8_AUTOMATED_UPDATE_ENABLED
-            valueFrom:
-              configMapKeyRef:
-                name: f8tenant
-                key: automated.update.enabled
+            value: ${f8_automated_update_enabled}
           image: ${IMAGE}:${IMAGE_TAG}
           imagePullPolicy: Always
           name: f8tenant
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-tenant/blob/872f0e87a1f3be3d1d26294c606aa2d2940eaaa2/.ike-prow/test-keeper.yaml).

OpenShiftTemplate.yml
```diff
@@ -184,3 +181,5 @@ parameters:
   value: quay.io/openshiftio/rhel-fabric8-services-fabric8-tenant
 - name: IMAGE_TAG
   value: latest
+- name: f8_automated_update_enabled
```
nurali-techie: don't you want to have upper case `F8_AUTOMATED_UPDATE_ENABLED` as in case of other parameters.

MatousJobanek: I followed the same approach as is already used in UI repo: https://github.com/fabric8-ui/fabric8-ui/blob/master/openshift/fabric8-ui.app.yaml#L171 https://github.com/openshiftio/saas-openshiftio/blob/master/dsaas-services/fabric8-ui.yaml#L11

xcoulon: well, I would have preferred that we remain consistent within this template, hence using `AUTOMATED_UPDATE_ENABLED` (no `f8_` prefix and uppercased) would have been better, IMO.

Issue #706, opened 2018-12-14
Create an endpoint for triggering cluster-wide tenant updates


MatousJobanek:
the endpoint called `/api/update` should:

* be limited for specific SA `fabric8-tenant-update`
* be able to start the update process (`POST`)
* be able to stop (`DELETE`)
* return information about last/ongoing run (`GET`)

PR #707, opened 2018-12-14
feat(#706): create /api/update endpoint to manage cluster-wide tenants updates

MatousJobanek:
This PR brings a new endpoint `/api/update` where:

* `POST` is for triggering the updates. It is possible to limit the update to one cluster (eg. `https://console.starter-us-east-2a.openshift.com`) or to one environment type (eg. `jenkins`). When such a limit is used, then when the update is finished, then the status is set to `incomplete`
* `GET` returns information about last/ongoing update.
* `DELETE` stops the process of doing updates and sets the status to `killed`

All endpoints are restricted to service account with name `fabric8-tenant-update`

As part of this PR I also added some additional logging

TODO:
- [x] add test for `/api/update` endpoint
- [x] add test for limiting to specific env type/cluster

Fixes: #706


controller/update.go
```diff
@@ -0,0 +1,126 @@
...
+// Show runs the show action.
+func (c *UpdateController) Show(ctx *app.ShowUpdateContext) error {
+	if !commonauth.IsSpecificServiceAccount(ctx, fabric8TenantUpdateSA) {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("Wrong token"))
+	}
+
+	var tenantsUpdate *update.TenantsUpdate
+	err := update.Transaction(c.db, func(tx *gorm.DB) error {
+		var err error
+		tenantsUpdate, err = update.NewRepository(tx).GetTenantsUpdate()
```
alexeykazakov: We don't have to execute read only operations in transactions.

design/update.go
```diff
@@ -0,0 +1,68 @@
...
+var _ = a.Resource("update", func() {
+	a.BasePath("/api/update")
+	a.Action("start", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.POST(""),
+		)
+		a.Params(func() {
+			a.Param("cluster_url", d.String, "the URL of the OSO cluster the update should be limited to")
+			a.Param("env_type", d.String, "environment type the update should be executed for")
```
alexeykazakov: Maybe it's worth adding a.Enum(...) to list all valid type here?

controller/update_test.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/fabric8-services/fabric8-tenant/controller"
+	"github.com/fabric8-services/fabric8-tenant/environment"
+	"github.com/fabric8-services/fabric8-tenant/tenant"
+	"github.com/fabric8-services/fabric8-tenant/test"
+	"github.com/fabric8-services/fabric8-tenant/test/doubles"
+	"github.com/fabric8-services/fabric8-tenant/test/gormsupport"
+	tf "github.com/fabric8-services/fabric8-tenant/test/testfixture"
+	"github.com/fabric8-services/fabric8-tenant/test/update"
+	"github.com/fabric8-services/fabric8-tenant/update"
+	"github.com/fabric8-services/fabric8-wit/ptr"
```
alexeykazakov: you can use ptr from common here.

design/update.go
MatousJobanek: good point  - added here 1746454

controller/update_test.go
MatousJobanek: slipped out of my notice, thanks - fixed 831bfcf

controller/update.go
MatousJobanek: In this particular case not, but in the rest of the cases, I need to read it in a transaction with the locked table. Not sure if there is some easy way to call the method without using a transaction when I need it in the other cases

MatousJobanek: wait, I can actually do this: b7cf997

controller/update.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/fabric8-services/fabric8-tenant/environment"
+	"github.com/fabric8-services/fabric8-tenant/jsonapi"
+	"github.com/fabric8-services/fabric8-tenant/openshift"
+	"github.com/fabric8-services/fabric8-tenant/update"
+	"github.com/fabric8-services/fabric8-wit/ptr"
+	"github.com/goadesign/goa"
+	"github.com/jinzhu/gorm"
+)
+
+var fabric8TenantUpdateSA = "fabric8-tenant-update"
```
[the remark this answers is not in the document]
MatousJobanek: sent PR https://github.com/fabric8-services/fabric8-common/pull/83

MatousJobanek:
> Looks good overall. But need to fix go fmt ;)

Actually, I formatted before i pushed, but with the wrong version of go :-/
fixed 6d40c58 
Thanks for the review

design/update.go
MatousJobanek: and here 8ac68a5 
the original commit is adding an enum only for status

MatousJobanek:
[test]

Issue #708, opened 2018-12-14
Wait for finishing tenant reset/update and show errors


MatousJobanek:
https://github.com/openshiftio/openshift.io/issues/4630

PR #709, opened 2018-12-14
fix(#708): wait for finishing tenant reset/update and return errors

MatousJobanek:
fixes: #708


MatousJobanek:
> Code wise look good but it seems that our tenant controller is covered by minishift tests only? We will need to cover it by CI tests too at some point (would need to mock OpenShift service for that).

unfortunately yes, it is covered only by minishift because it executes `oc` command that is hard to mock. The new code that is proposed here https://github.com/fabric8-services/fabric8-tenant/pull/673 adds much integration tests that are executed in CI

Issue #720, opened 2019-01-04
Make the cluster-wide update slower


MatousJobanek:
The current implementation is too fast - runs 10 tenant updates concurrently and does not wait between the particular tenant updates. This could break prod in case of running the updates for `jenkins` namespace (could run out of resources).

PR #721, opened 2019-01-04
issue(#720): make cluster-wide tenants update slower

MatousJobanek:
This PR makes the tenants update slower to avoid capacity issues while running it in prod.
* do not run the tenant updates concurrently
* wait for 4 seconds after a single tenant update

This PR is a result of a discussion with @pbergene

Fixes: #720

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -48,6 +48,7 @@ const (
 	varEnvironment                     = "environment"
 	varSentryDSN                       = "sentry.dsn"
 	varAutomatedUpdateRetrySleep       = "automated.update.retry.sleep"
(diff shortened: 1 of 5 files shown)


MatousJobanek:
> Do we run all updates for all clusters in a sequence now? Or we still do the updates for different clusters in parallel? 

Yes, we do run it in sequence. For the first runs I'm going to trigger the update for every cluster separately anyway (just to make it as save as possible). 

> Asking because we should run updates in a sequence within one cluster but it's fine to run them in parallel in different clusters, isn't it?

I think that it should be fine running them in parallel - that is my optimization task for the future when we know that this way of running updates is working fine :-)

Issue #724, opened 2019-01-09
don't recreate when update fails


MatousJobanek:
when the `oc apply` update fails, then don't recreate namespaces and just wait and try to run the update one more time

PR #725, opened 2019-01-09
fix(#724): don't recreate when update fails

MatousJobanek:
when the update fails with one or more errors, then don't recreate and just wait for 10s and try to update it one more time.

fixes: #724


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-tenant/blob/a73945bd492c90d694f9afd701080e0d8180429a/.ike-prow/test-keeper.yaml).

PR #728, opened 2019-01-16
move transaction support to separated package-to be usable in future

MatousJobanek:
moving transaction logic to a separated package so it is usable in the future

### DIFF
--- a/update/repository.go
+++ b/update/repository.go
@@ -3,6 +3,7 @@ package update
 import (
 	"database/sql/driver"
 	"fmt"
+	"github.com/fabric8-services/fabric8-tenant/dbsupport"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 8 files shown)


update/repository.go
```diff
@@ -150,3 +151,18 @@ func (r *GormRepository) Stop() error {
 	}
 	return nil
 }
+
+const TenantsUpdateAdvisoryLockID = 4242
+
+func lock(do func(repo Repository) error) dbsupport.LockAndDo {
```
xcoulon: why don't you move this `lock` function in the new `dbsupport` package? it seems like a common function that can be reused accros multiple repositories, doesn't it?

MatousJobanek: done 32ee892

xcoulon: great, thanks!

Issue #731, opened 2019-01-21
Return number of outdated tenants


MatousJobanek:
When GET `api/update` is called, then it should also return a number of outdated tenants( that are going to be updated). it should support limitation to both env type and master URL

PR #732, opened 2019-01-21
feat(#731): return number of outdated tenants

MatousJobanek:
When GET to `api/update` is called, then the response (apart from the information about an ongoing update) contains also a number of outdated tenants (that are going to be updated) `to-update`. 
It supports limitation to both env type and master URL.

Fixes: #731

### DIFF
--- a/tenant/repository.go
+++ b/tenant/repository.go
@@ -26,6 +26,7 @@ type Service interface {
 	NamespaceExists(nsName string) (bool, error)
 	ExistsWithNsBaseName(nsBaseName string) (bool, error)
 	GetTenantsToUpdate(typeWithVersion map[environment.Type]string, count int, commit string, masterURL string) ([]*Tenant, error)
+	GetNumberOfOutdatedTenants(typeWithVersion map[environment.Type]string, commit string, masterURL string) (int, error)
 }
 
 func NewDBService(db *gorm.DB) Service {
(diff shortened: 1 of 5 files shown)


tenant/repository.go
```diff
@@ -127,6 +128,19 @@ func (s DBService) GetNamespaces(tenantID uuid.UUID) ([]*Namespace, error) {
...
+}
+
+func (s *DBService) GetNumberOfOutdatedTenants(typeWithVersion map[environment.Type]string, commit string, masterURL string) (int, error) {
+	var count int
+	err := s.createGetOutdatedTenantsQuery(typeWithVersion, commit, masterURL).Count(&count).Error
+
+	return count, err
+}
+
+func (s *DBService) createGetOutdatedTenantsQuery(typeWithVersion map[environment.Type]string, commit string, masterURL string) *gorm.DB {
```
xcoulon: this func could be named `newGetOutdatedTenantsQuery` instead

tenant/repository_test.go
```diff
@@ -322,6 +322,93 @@ func (s *TenantServiceTestSuite) TestGetSubsetOfTenantsThatMatchesRequiredCluste
 	})
 }
 
+func (s *TenantServiceTestSuite) TestGetNumberOfAllOutdatedTenants() {
+	// given
+	configuration.Commit = "123abc"
+	testdoubles.SetTemplateVersions()
+	tf.FillDB(s.T(), s.DB, tf.AddTenants(3), tf.AddDefaultNamespaces().State(tenant.Ready).Outdated())
+	svc := tenant.NewDBService(s.DB)
```
xcoulon: the name for `NewDBService` is not usual: in `auth` and `cluster` services, we have a `service` layer and a `repository` layer (this later uses GORM to run SQL queries). So `DBService` looks like a mix of both

tenant/repository.go
MatousJobanek: good point, renamed 37a916f

tenant/repository_test.go
MatousJobanek: I know, I don't like the name from the very first moment. I don't like the name of the package either.  It's one of the things I wanted to change as part of a rewrite of this DB representation.

I could change it in the [big core PR](https://github.com/fabric8-services/fabric8-tenant/pull/673), however, there is already `Repository` introduced so I could name it `Service`. But in the combination with the package name it would result in `tenant.NewService(s.DB)` and (from my point of view) this looks weird - the name `tenant.Service` doesn't say anything about DB or repo. That's why I left the original name there for now.

xcoulon: @MatousJobanek ok let's revisit this once those 2 PRs have been merged.

Issue #729, opened 2019-01-18
Introduce lazy initialization


MatousJobanek:
Tenant service should not wait for the first POST call but should initialize namespaces when it receives first GET call containing user's token

nurali-techie:
@MatousJobanek can you plz detailed out the motivation behind this changes OR some issue encounter which need this changes.

MatousJobanek:
@nurali-techie the motivation is to be ready for removing POST calls to tenant service when a user logs in - it's mentioned in the issue you have created if I'm not mistaken: https://github.com/fabric8-services/fabric8-tenant/issues/746
Anyway, the original motivation for removing the POST call is an error that tenant service sometimes gets for new users from auth service - it sometimes fails that the user still doesn't have OS token assigned so it is not possible to provision the namespaces

nurali-techie:
@MatousJobanek thanks, as we are still discussing #746 I was not expecting any changes wrt to the issue at the same time I understand that #729 may impact #746 so I thought to check what's the original motivation for #729.

PR #742, opened 2019-01-29
feat(#729): initialize tenant when GET call is received and don't wait for POST

MatousJobanek:
* tenant initializes namespaces (and creates DB records) not only for `POST` call but also for `GET` call that is received with a user token. 
* checks for the existence of all namespaces - in other words - when one namespace is missing for a particular tenant, then it initializes it

Fixes: #729


MatousJobanek:
[test]

controller/tenant.go
```diff
@@ -337,6 +298,74 @@ func listNames(namespaces []*tenant.Namespace) []string {
 	return names
 }
 
+func (c *TenantController) getOrInitTenant(ctx context.Context, user *auth.User) (*tenant.Tenant, error) {
+	var dbTenant *tenant.Tenant
+	var err error
+	tenantRepository := c.tenantService.NewTenantRepository(user.ID)
```
xcoulon: why do you pass the `user.ID` in the tenant repository constructor? shouldn't it be user-agnostic (hence, pass the `user.ID` as an arg when calling the methods. eg: `repository.Exists(user.ID)`)

controller/tenant_parallel_test.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func TestTenantControllerInParallel(t *testing.T) {
+	suite.Run(t, &TenantControllerParallelTestSuite{DBTestSuite: gormsupport.NewDBTestSuite("../config.yaml")})
+}
+
+type whenInParallel func(ctx context.Context, service *goa.Service, ctrl app.TenantController)
+
+func (s *TenantControllerParallelTestSuite) TestShowTenantOKWhenNoTenantExistsInParallelForOneUser() {
+	s.verifyTenantCreationWhenNoTenantExistsInParallelForOneUser(
```
xcoulon: this is a long method name 😂

controller/tenant.go
```diff
@@ -246,7 +190,24 @@ func (c *TenantController) Show(ctx *app.ShowTenantContext) error {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 
-	return ctx.OK(&app.TenantSingle{Data: convertTenant(ctx, tenant, namespaces, c.clusterService.GetCluster)})
+	// check if any environment type is missing - should be provisioned
+	missing, _ := filterMissingAndExisting(namespaces)
```
xcoulon: do you have a test for this `filterMissingAndExisting` func?

controller/tenant.go
```diff
@@ -337,6 +298,74 @@ func listNames(namespaces []*tenant.Namespace) []string {
...
+				"err": err,
+			}, "unable to store tenant configuration")
+			return nil, err
+		}
+	}
+	return dbTenant, nil
+}
+
+func (c *TenantController) createNamespaces(ctx context.Context, envTypes []environment.Type, user *auth.User, dbTenant *tenant.Tenant) error {
+	clusterNsMapping, err := c.clusterService.GetUserClusterForType(ctx, user)
```
xcoulon: why is this method called `GetUserClusterForType`? what is the `type` here?

controller/tenant.go
```diff
@@ -337,6 +298,74 @@ func listNames(namespaces []*tenant.Namespace) []string {
...
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err":         err,
+			"tenant":      user.ID,
+			"cluster_url": *user.UserData.Cluster,
```
xcoulon: are you sure that `*user.UserData.Cluster` won't panic?

controller/tenant.go
MatousJobanek: Because the `TenantRepository` is specific for one tenant. In one request I'm dealing with one tenant and this helps me to minimize the number of parameters passed to other methods/objects - I don't have to pass DB service along with the tenant id because it is already contained within the repository: https://github.com/fabric8-services/fabric8-tenant/blob/develop/tenant/repository.go#L138

controller/tenant_parallel_test.go
MatousJobanek: I know, it's just a replacement of the test and it explains what it does... Do you want me to change it? Proposal?

controller/tenant.go
xcoulon: yes, that's what I understood, but it feels unusual to construct the repository with the tenant associated to it. My point is that it does not follow our pattern for services and repositories.

controller/tenant.go
MatousJobanek: https://github.com/fabric8-services/fabric8-tenant/blob/develop/cluster/service.go#L34
I could maybe change it to `GetUserClusterForEnvType`

controller/tenant_parallel_test.go
xcoulon: nope, it's ok. The name is self-descriptive ;)

controller/tenant.go
MatousJobanek: Yes, I am. I both cases the user info has to be already fetched (otherwise it would fail earlier)

controller/tenant.go
xcoulon: but you only pass the context and user as args here, so why `ForType` or `ForEnvType` ?

controller/tenant.go
xcoulon: ok, good 👍

controller/tenant.go
MatousJobanek: only indirect tests - not direct
see: 
* when no namespace exists
`TestShowTenantOKWhenNoTenantExists`
`TestSetupTenantOKWhenNoTenantExists`

* when some namespaces exist 
`TestShowTenantWhenSomeNamespacesAreMissing`
`TestSetupTenantOKWhenAlreadyExists`

* when all namespace exist
`TestSetupConflictFailure`
`TestShowTenant`


in https://github.com/fabric8-services/fabric8-tenant/pull/742/files#diff-f9e0701950f6288288104f0f34160817

controller/tenant.go
MatousJobanek: For me it makes sense - it represents a repository for one tenant entity comparing to the `DBService` that represents a service for all other cases (not related to one tenant). 
But I'm open to any proposal ;-).
we already had some discussions about it:
https://github.com/fabric8-services/fabric8-tenant/pull/673#discussion_r248610738
https://github.com/fabric8-services/fabric8-tenant/pull/732#discussion_r249411930

I have a big problem with these namings so I would welcome any help :-)

controller/tenant.go
MatousJobanek: it returns a function `ForType` that returns cluster for the given environment type. To point here is that it was a preparation for a multi-cluster environment (when one env type can be associated with a different cluster), that's why such a mapping. In this case, it just returns the same cluster (the user cluster) for all env types. This function is then used in a generic way.

MatousJobanek:
Closing this issue as it is not relevant anymore. In addition, as there is a Get call before the first Post call, then it would be too early to start provisioning for the first Get call

Issue #730, opened 2019-01-18
Reflect removed objects in templates


MatousJobanek:
When an object is removed from some tenant template, then during the next update the tenant service should remove the corresponding object from all tenant namespaces.
To do that, it has to compare the last template that was used for the creation/update of a particular namespace and compare it with the current one (if it contains some extra objects that were removed)

PR #741, opened 2019-01-29
feat(#686): add/remove user-edit rolebinding in che based on toggle

MatousJobanek:
based on the information whether the user has `che.edit.rights` toggle enabled it adds or removes the `user-edit` rolebinding for the che namespace. The body that is sent to OS (when doing `POST`) is: 
```yml
kind: RoleBinding
metadata:
  name: user-edit
  namespace: ${ns-base-name}-che
roleRef:
  name: edit
subjects:
- kind: User
  name: ${os-username}
userNames:
- ${os-username}
```

Fixes:  #686

PR #745, opened 2019-02-05
 fix(#730): delete objects that are removed from templates 

MatousJobanek:
The code to be reviewed is in this commit 8630aa6

* when an object is removed from any of the templates, then during the next update it removes the corresponding object from an associated namespace in OS

* the deletion is done by comparing the content of the templates that were used for the update/provisioning of the namespace last time. It takes the version of the namespace and because it represents a commit in tenant repo, then it downloads the content of the template from the corresponding blob. Then it just compares the objects - combination of kind/name/namespace 

* introduces a caching mechanism to minimize the number of downloads of the previous version of the templates. The cache contains a record for a particular version and for a particular environment type. The actual target of the cache is list of removed objects   

is blocked by #741 as it is built on top of it

Fixes: #730

### DIFF
--- a/openshift/types.go
+++ b/openshift/types.go
@@ -121,6 +121,9 @@ type CheNamespaceTypeService struct {
 }
 
 func (t *CheNamespaceTypeService) AdditionalObject() (environment.Object, bool) {
+	if t.context.requestCtx == nil {
+		return environment.Object{}, true
+	}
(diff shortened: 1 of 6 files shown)


fabric8cd:
This PR changes one or more of the template files. These changes are now available for testing: Launch in [OpenShift.io](https://openshift.io/_profile/_tenant?templatesRepoBlob=a12152c6ce909f5f95cf758a95e75812d9428365) and click the update tenant button

PR #753, opened 2019-02-12
fix: delete objects in correct order & use merge-patch type for updates

MatousJobanek:
When doing clean action, the DeploymentConfig should be removed before ReplicationController and this one before Pods.


MatousJobanek:
[test]

PR #755, opened 2019-02-15
fix: returns conflict on duplicate key errors&set params in updater

MatousJobanek:
* When there is an error containing `pq: duplicate key value violates unique constraint` returned on the creation of tenant entity in DB, then it means that there is already one created so it should return a conflict (this happens when there are two POST call at the same time)
* construct TenantUpdater in main.go correctly to not panic

### DIFF
--- a/controller/tenant.go
+++ b/controller/tenant.go
@@ -16,6 +16,7 @@ import (
 	"github.com/goadesign/goa"
 	errs "github.com/pkg/errors"
 	"github.com/satori/go.uuid"
+	"strings"
 )
 
 // TenantController implements the tenant resource.
@@ -175,6 +176,9 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
 		}
 		err = tenantRepository.CreateTenant(dbTenant)
 		if err != nil {
+			if strings.Contains(err.Error(), "pq: duplicate key value violates unique constraint") {
+				return ctx.Conflict()
+			}
 			log.Error(ctx, map[string]interface{}{
 				"err": err,
 			}, "unable to store tenant configuration")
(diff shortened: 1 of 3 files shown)


controller/tenant.go
```diff
@@ -175,6 +176,9 @@ func (c *TenantController) Setup(ctx *app.SetupTenantContext) error {
 		}
 		err = tenantRepository.CreateTenant(dbTenant)
 		if err != nil {
+			if strings.Contains(err.Error(), "pq: duplicate key value violates unique constraint") {
```
nurali-techie: Don't we have better way to check if there is such error rather checking error_msg_string.
OR
Can we first call something like below ??
```
if repo.GetTenant(tenantId) != nil {
  return ctx.Conflic()
} else {
  repo.CreateTenant()
}
```

MatousJobanek: > Don't we have better way to check if there is such error rather checking error_msg_string.

I would love to have, but gorm doesn't provide such an error kind: https://github.com/jinzhu/gorm/blob/master/errors.go

> Can we first call something like below ??

I check for the existence already https://github.com/fabric8-services/fabric8-tenant/pull/755/files/32679607ef9ce58f269236b6dcc6b36ec48d7d96#diff-dfb0ecdc519ac765cb68432d078cc1d0R143 but this doesn't work for two POST calls that run at the same time concurrently - in that case, I would need to lock the table

nurali-techie:
@MatousJobanek when I was debugging tenant_lazy_init, I came across code which I guess make concurrent call to `POST /api/tenant` during user login from openshift.io.

During login, call comes to wit_service user#Show() [code_link](https://github.com/fabric8-services/fabric8-wit/blob/master/controller/user.go#L54) which indirectly make TWO POST call to InitTenant.
First call - [code_link](https://github.com/fabric8-services/fabric8-wit/blob/master/controller/user.go#L59) this f/w call to auth_service which call async`POST /api/tenant`.
Second call - [code_link](https://github.com/fabric8-services/fabric8-wit/blob/master/controller/user.go#L81) which call async `POSt /api/tenant`

MatousJobanek:
[test]

Issue #757, opened 2019-02-18
Run tenants update for clusters in parallel


MatousJobanek:
When no `cluster_url` is set  as a limit for the cluster-wide update, then the updater should be run in parallel for the list of all clusters that contain outdated tenants

PR #758, opened 2019-02-18
feat(#757): update clusters in parallel

MatousJobanek:
When no `cluster_url` is set as a limit for the cluster-wide update, then the updater is executed for the list of all clusters that contain outdated tenants in parallel.

fixes: #757

### DIFF
--- a/tenant/repository.go
+++ b/tenant/repository.go
@@ -19,6 +19,7 @@ type Service interface {
 	NamespaceExists(nsName string) (bool, error)
 	ExistsWithNsBaseName(nsBaseName string) (bool, error)
 	GetTenantsToUpdate(typeWithVersion map[environment.Type]string, count int, commit string, masterURL string) ([]*Tenant, error)
+	GetClustersToUpdate(typeWithVersion map[environment.Type]string, commit string) ([]string, error)
 	GetNumberOfOutdatedTenants(typeWithVersion map[environment.Type]string, commit string, masterURL string) (int, error)
 }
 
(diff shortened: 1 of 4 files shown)


tenant/repository.go
```diff
@@ -89,15 +90,29 @@ func (s *DBService) GetTenantsToUpdate(typeWithVersion map[environment.Type]stri
...
+func (s *DBService) GetClustersToUpdate(typeWithVersion map[environment.Type]string, commit string) ([]string, error) {
+	var namespaces []*Namespace
+	err := s.newGetOutdatedNamespacesQuery(typeWithVersion, "master_url", commit, "").Scan(&namespaces).Error
+	if err != nil {
+		return nil, err
+	}
+	clusters := make([]string, 0, len(namespaces))
+	for _, ns := range namespaces {
+		clusters = append(clusters, ns.MasterURL)
+	}
```
alexeykazakov: It will actually return masterURLs from all namesapces, won't it? I mean there will be a lot of duplicates.
If we have 100 namespaces with east-1 masterURL and 200 with west-2 then we will get 300 "clusters", won't we?

MatousJobanek: It should not - there's `GROUP BY` used in the query https://github.com/fabric8-services/fabric8-tenant/pull/758/files/76b8d1ec1cb8be0f8f99f27fbe4d120895373c90#diff-98a18397d2c4ed836ccd3f44b218549dR127
There is also a test that should detect it: https://github.com/fabric8-services/fabric8-tenant/pull/758/files/76b8d1ec1cb8be0f8f99f27fbe4d120895373c90#diff-09939f779f4b72015b5455cd34a0026cR181 There are added three tenants per cluster URL, which would end up in 9 records (but is checked that there are only three), if I'm not mistaken.

alexeykazakov: Ah, right. I missed that.

PR #763, opened 2019-03-15
feat(ODC-238): add tenant related monitoring metrics 

MatousJobanek:
adds monitoring metrics that record number of provisioned/updated/cleaned tenants.
Every record contains information whether the action was successful or failed. In addition, updated/cleaned records also contain namespace base name of the affected tenant.

Implements: [ODC-238](https://jira.coreos.com/browse/ODC-238)

Builds on top of https://github.com/fabric8-services/fabric8-tenant/pull/762, the actual commit that implements the metrics is this c4573fd

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -366,11 +366,12 @@
   version = "v1.0.0"
 
 [[projects]]
-  digest = "1:735956b5e94182229dfbdc5b401d7a71f20f54f1259dbc2969ec2b56bfbb34f2"
+  digest = "1:26663fafdea73a38075b07e8e9d82fc0056379d2be8bb4e13899e8fda7c7dd23"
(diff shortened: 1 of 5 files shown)


MatousJobanek:
> So, we just keep incrementing successful/failed provisions/updates/cleanups without resetting the numbers? How will the metrics look like in the real life?

I don't know, to be honest. I don't have any experience in this field but I expect that it should be possible to filter the counts by a hours/days/weeks in the resulting UI... 
I don't think that resetting is a good approach here - I went through all repos that use these metrics and no-one resets them. 
I'd say, lest see what it looks like and then (according to the results) I'll try to change it (use other metric or label) if necessary.
Maybe @nurali-techie or @dipak-pawar could shed more light on this (I think that both of them have experience with using metrics)

PR #764, opened 2019-03-25
disable check of unprovisioned namespaces for already provisioned tenant

MatousJobanek:
temporary workaround for https://github.com/redhat-developer/rh-che/issues/1296#issuecomment-475149410

It disables the check of unprovisioned namespaces for already provisioned tenants. This allows us to have an account with only two namespaces (che and user) provisioned

### DIFF
--- a/controller/tenant_test.go
+++ b/controller/tenant_test.go
@@ -136,7 +136,7 @@ func (s *TenantControllerTestSuite) TestSetupTenantOKWhenNoTenantExistsInParalle
 
 			// then
 			if err != nil {
-				assert.Equal(s.T(), err, test.IsOfType(setupCtx.Conflict()))
+				test.AssertError(s.T(), err, test.HasMessageContaining("conflict"))
 			}
 		}()
 	}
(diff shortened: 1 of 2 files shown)


controller/tenant_test.go
```diff
@@ -194,11 +194,11 @@ func (s *TenantControllerTestSuite) TestSetupTenantOKWhenNoTenantExistsInParalle
 				run.Wait()
 
 				// when
-				err = ctrl.Setup(setupCtx)
+				ctrl.Setup(setupCtx)
```
nurali-techie: LHS is missing and sub-sequent checks will be wrong.

MatousJobanek: good catch, thanks. fixed 8d16c1c

controller/tenant_test.go
```diff
@@ -221,18 +223,18 @@ func (s *TenantControllerTestSuite) TestSetupTenantOKWhenAlreadyExists() {
...
+	apptest.SetupTenantConflict(s.T(), testdoubles.CreateAndMockUserAndToken(s.T(), id.String(), false), svc, ctrl)
 	// then
-	totalNumber := testdoubles.ExpectedNumberOfCallsWhenPost(s.T(), config)
-	cheObjects := testdoubles.SingleTemplatesObjectsWithDefaults(s.T(), config, environment.TypeChe)
-	numberOfGetChecksForChe := testdoubles.NumberOfGetChecks(cheObjects)
-	assert.Equal(s.T(), totalNumber-(len(cheObjects)+numberOfGetChecksForChe+1), calls)
+	//totalNumber := testdoubles.ExpectedNumberOfCallsWhenPost(s.T(), config)
+	//cheObjects := testdoubles.SingleTemplatesObjectsWithDefaults(s.T(), config, environment.TypeChe)
+	//numberOfGetChecksForChe := testdoubles.NumberOfGetChecks(cheObjects)
+	//assert.Equal(s.T(), totalNumber-(len(cheObjects)+numberOfGetChecksForChe+1), calls)
```
nurali-techie: No checks after call is made.  Is there anything we can check here ??

PR #769, opened 2019-05-02
fix: when tenant is deleted,remove namespaces of unsupported types from DB

MatousJobanek:
When a tenant is going to be deleted, then remove also DB records of namespaces of old unsupported namespaces (run, stage, jenkins) to not fail when checking if all namespaces were removed.

### DIFF
--- a/controller/tenants_test.go
+++ b/controller/tenants_test.go
@@ -190,6 +190,34 @@ func (s *TenantsControllerTestSuite) TestSuccessfullyDeleteTenants() {
 			HasNoNamespace()
 	})
 
+	s.T().Run("ok when unsupported namespaces exist in DB", func(t *testing.T) {
+		// given
+		defer gock.OffAll()
+		testdoubles.MockCommunicationWithAuth(test.ClusterURL)
+		gock.New(test.ClusterURL).
+			Delete("/oapi/v1/projects/foo-che").
+			SetMatcher(test.ExpectRequest(test.HasJWTWithSub("devtools-sre"))).
+			Reply(200).
+			BodyString(`{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Success"}`)
+		gock.New(test.ClusterURL).
+			Delete("/oapi/v1/projects/foo").
+			SetMatcher(test.ExpectRequest(test.HasJWTWithSub("devtools-sre"))).
+			Reply(200).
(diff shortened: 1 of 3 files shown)


openshift/action_test.go
```diff
@@ -562,6 +562,21 @@ func (s *ActionTestSuite) TestDeleteAction() {
 		assertion.AssertTenant(t, repo).DoesNotExist()
 	})
 
+	s.T().Run("ManageAndUpdateResults should delete also unsupported entities", func(t *testing.T) {
```
xcoulon: did you mean "unsupported namespaces" ?

controller/tenants_test.go
```diff
@@ -190,6 +190,34 @@ func (s *TenantsControllerTestSuite) TestSuccessfullyDeleteTenants() {
...
+			SetMatcher(test.ExpectRequest(test.HasJWTWithSub("devtools-sre"))).
+			Reply(200).
+			BodyString(`{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Success"}`)
+		gock.New(test.ClusterURL).
+			Delete("/oapi/v1/projects/foo").
+			SetMatcher(test.ExpectRequest(test.HasJWTWithSub("devtools-sre"))).
+			Reply(200).
+			BodyString(`{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Success"}`)
+
+		fxt := tf.FillDB(t, s.DB, tf.AddSpecificTenants(tf.SingleWithName("foo")),
```
xcoulon: ok, so in this test, you fill the DB with records for some `user`, `che`, `jenkins`, `stage` and `run` tenants but only `user` and `che` will be deleted? All other namespaces/tenants are assumed to be already deleted (ie, `jenkins`, `stage` and `run`)? And at the end of the test, all records were removed from the DB?

openshift/action_test.go
MatousJobanek: yup :-) fixed in 4a7f712

controller/tenants_test.go
MatousJobanek: Yes, exactly :-)

xcoulon: ok, thanks for clarifying ;)

openshift/action_test.go
xcoulon: thanks ;)

PR #776, opened 2019-09-16
Add resources.requests to OpenShift template

alexeykazakov:


alien-ike:
### Ike Plugins (test-keeper)

Thank you @alexeykazakov for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-tenant/blob/a08589098965fb5fafe1cb397d9b17946a3edcd4/.ike-prow/test-keeper.yaml).

OpenShiftTemplate.yml
```diff
@@ -161,6 +161,9 @@ objects:
             successThreshold: 1
             timeoutSeconds: 1
           resources:
+            requests:
```
xcoulon: In Jaime's mail, there was the following recommendation:

> Limits: we recommend that the number represents an abnormal amount of
resource usage that could be caused by a memory leak, etc. We would
usually expect this value to be not too far apart from the
corresponding requests value, something around 2x

here the limits/requests ratio is much higher than 2x. (it's 40x for the CPU and 20x for the memory). How did you choose those values?

alexeykazakov: We had these limit numbers for auth service for two+ years.
The request numbers (which is what this PR is about) I took from the monitoring data.

My goal here is to comply with the new requirements and not to break anything in the same time.
So, the request numbers IMO are reasonable. The limits are probably too high but it's better than not having them at all (and we didn't have CPU limits before at all). And we are on the safe side with these numbers IMO.

I don't want to spend too much time on fine tuning the numbers.

xcoulon: ok, thanks for your feedback, @alexeykazakov

PR #781, opened 2020-01-08
fix(CRT-355): use all vars as strings in all templates

MatousJobanek:
use all vars as strings in all templates because not only annotations have a problem with a username containing only numbers, but also RoleBindings. Yes, I'm stupid that I didn't realize it before :-/


MatousJobanek:
[test]

PR #785, opened 2020-02-27
chore(CRT-473): disable sentry

MatousJobanek:


MatousJobanek:
> Shouldn't we remove these sentry calls completly?

Yes, good catch - thanks a lot. Done in df4dd86

alexeykazakov:
Sorry, but it's still not enough... :) In some places sentry package from common is used too. For example https://github.com/fabric8-services/fabric8-tenant/blob/9c2ea27266be4dd7267a9c605b1ad7d56c601287/controller/convert.go#L22
We need to remove these calls too.

MatousJobanek:
> Sorry, but it's still not enough... :) In some places sentry package from common is used too. For example 

no, these calls use the internal sentry package (inside of the tenant service). To make this change easier I modified only directly the method so all the calls can keep the same:
https://github.com/fabric8-services/fabric8-tenant/pull/785/files#diff-81d240a19b16a9de576826a538b41874R46-R51

to make sure that the sentry is not used anywhere I comented out the whole implementations which caused that the sentry is not used in the repo anymore: 
https://github.com/fabric8-services/fabric8-tenant/pull/785/files#diff-bd247e83efc3c45ae9e8c47233249f18L649

alexeykazakov:
Sorry. Wrong link. I meant calls like this https://github.com/fabric8-services/fabric8-tenant/blob/9c2ea27266be4dd7267a9c605b1ad7d56c601287/jsonapi/jsonapi_utility.go#L153
But it seems to be the only use of sentry from common and you have already removed it. So, looks good!

